PR #2, opened 2016-09-19
Add missing PureRender dependency

tmcw:


lukasmartinelli:
> tmcw

Thanks for the missing dependency.

PR #15, opened 2016-11-09
Serve the editor using container

chrissng:
- Added a Dockerfile to allow the editor to be served using docker containers
- Updated README with docker instructions

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,15 @@
+FROM nodesource/xenial:6.1.0
+
+EXPOSE 8888
+
+ENV HOME /maputnik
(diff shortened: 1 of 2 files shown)


lukasmartinelli:
Very nice. Thanks @chrissng for solving this.
In the future I hope people can use this binary here https://github.com/maputnik/desktop together with the bundled assets (so it is easy for Windows folks as well).

chrissng:
@lukasmartinelli thanks for approving and merging my PR! I missed the kickstarter campaign so I thought I could help by making a small contribution in some useful way.

PR #36, opened 2016-12-28
Upgraded reference to mapbox-gl to v0.29.0 to follow their latest webpack recommendations and support Windows dev. This is addressed in mapbox/mapbox-gl-js#3724

PetersonGIS:


lukasmartinelli:
Oh wow thanks a lot Gretchen (and husband). :pray: 

This is really the best way to go. Avoided using the prebuilt Mapbox GL so far in the hope to make the final JavaScript bundle lighter (by sharing dependencies) but this solution works cross platform and is more maintainable in the future. Thanks for digging into it.

I also provide the public HTML bundle on the[ releases page ](https://github.com/maputnik/editor/releases)(for unzipping and opening the HTML file) for people who don't have Node installed.

PetersonGIS:
Great, glad to help!

PR #55, opened 2017-01-06
Format style on download

jirik:
This should clear commit history of styles. If style is not formatted, diff may look like this:
https://github.com/openmaptiles/osm-bright-gl-style/commit/c514b0ce7fa5335f3b0e8334aed0ea0d05f8d5cc


lukasmartinelli:
Thanks Jiri!!

PR #71, opened 2017-01-12
Reordered and renamed groups

muesliq:
Purpose: Match structure and wording with Mapbox GL style spec.


lukasmartinelli:
Ah it's better than before for sure. I merge it.

Issue #70, opened 2017-01-12
Cannot Change Value for Boolean Values in Zoom Level Functions


lukasmartinelli:
![screenshot from 2017-01-12 14-07-07](https://cloud.githubusercontent.com/assets/1288339/21890922/6d3e3524-d8d0-11e6-9058-76a9f35e4eaf.png)

lukasmartinelli:
This works now. Seems to have been fixed by @jirik 
![screenshot from 2017-01-23 09-36-02](https://cloud.githubusercontent.com/assets/1288339/22196826/6ccb6526-e14f-11e6-9f52-003c9b66ab62.png)

PR #86, opened 2017-01-16
Gist preview & access token

jirik:
When saving to Gist, user can choose if he wants a preview. If so, he needs to add access token that can be generated for free from the link below the input. It is the same access token from style settings.

In this PR, I changed also behaviour of two general components:
fab004cdfe4de1446074fd60e50b1d775a8d26d7 I did this because it did not call onChange listener if new value was empty string.
3b7fb7ae75c88f22bd0489a00d6fd5261618ebee I did this because it CheckBox always showed "check" icon despite being true or false.

@lukasmartinelli please check two mentioned commits carefully.

### DIFF
--- a/src/components/Toolbar.jsx
+++ b/src/components/Toolbar.jsx
@@ -91,6 +91,7 @@ export default class Toolbar extends React.Component {
       />
       <ExportModal
         mapStyle={this.props.mapStyle}
+        onStyleChanged={this.props.onStyleChanged}
         isOpen={this.state.isOpen.export}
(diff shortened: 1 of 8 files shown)


lukasmartinelli:
Tweaked the CSS a bit. Otherwise great.

> @lukasmartinelli please check two mentioned commits carefully.

Thanks for fixing the behavior. It might be you also fixed #70 - since this works now. 

It's probably the best possible option right now. Compared to before it is a hurdle to publish the maps - even though the access token create link is necessary. The button before was just one click - and you got a map. Which was cool.

Issue #44, opened 2017-01-04
Color Picker Format and Alpha Values


lukasmartinelli:
The Alpha value slider does not work.

![screenshot from 2017-01-04 09-06-40](https://cloud.githubusercontent.com/assets/1288339/21635301/22bef3f4-d25d-11e6-9e53-11f6366b0c91.png)

muesliq:
Also: Color picker doesn't stay within a color format:

1. I click on a color field with **hsl** value.
2. The color picker opens showing the corresponding **HEX** value.
3. I click on a different color
4. The value in the color field is updated with the **rgba** value.

This is confusing. IMO formats should stay consistent (hsl in this case).

lukasmartinelli:
> This is confusing. IMO formats should stay consistent (hsl in this case).

:+1: I agree.

Right now I always request rgba from the color picker.
The main problem is that I don't get the chosen color state from the color picker. So I don't know that you have chosen a HSLA color in the picker - I only get the rgba value.

PR #101, opened 2017-03-07
Alpha value slider fix (#44)

orangemug:
I've fixed the alpha value slider in the color picker, the issue was the color module returned `alpha` rather than `a` which `react-color` was expecting.

This fixes issue #44


lukasmartinelli:
> I've fixed the alpha value slider in the color picker, the issue was the color module returned alpha rather than a which react-color was expecting.

Ohh this was one of the things I hated most.
Great you found that out - actually spent 0.5hr there and didn't figure it out.

PR #125, opened 2017-04-04
Initial webdriver test

orangemug:
This is an initial webdriver test so I can start to build out the UI tests. I replaced karma because it was being flakey on my machine. Also I wanted to use the WebDriver Protocol to do browser testing because I believe its more robust.

I also moved as much of the config into `./config/` as to tidy up the root directory a little.


lukasmartinelli:
> This is an initial webdriver test so I can start to build out the UI tests. I replaced karma because it was being flakey on my machine. Also I wanted to use the WebDriver Protocol to do browser testing because I believe its more robust.

Oh wow go for it. Also like the idea of moving the webpack configs into the `config` folder.

orangemug:
So I've also moved to testing all the latest patches on the CI systems so `4.x`, `5.x`, `6.x` and `7.x` (see <https://github.com/nodejs/LTS>)

I've had to add in a C compiler on the CI envs for odd number node.js versions, see <https://github.com/laverdet/node-fibers/issues/333#issuecomment-282312238>

> Odd-numbered versions of Node are early release so you will need to install MSVC and the rest of the dependencies listed on the node-gyp GitHub page. Alternatively you can use node v6, or wait for node v8 this April.

This makes it slightly harder to install / build but **only** for odd node.js versions which I think is fine.

orangemug:
Ok, so after a lot of fighting with the CI envs the status of this PR is that Windows tests are failing on node v5 (see <https://ci.appveyor.com/project/lukasmartinelli/editor/build/1.0.191>). Although I believe it can be fixed I don't have time to look into it and I don't have a windows box to test locally. I suggest we just remove support for node v5 as its not in the supported list, see <https://github.com/nodejs/LTS>

orangemug:
Ok this PR is finally ready to review/merge, tests passing with node v5 removed on windows (see comment above)

lukasmartinelli:
> Ok this PR is finally ready to review/merge, tests passing with node v5 removed on windows (see comment above)

Looked over it and it looks good to me! Tests are green - perfect world.
This will prevent changes from breaking the entire application which is good.

orangemug:
More tests to come soon. What I want next is a tiny vector tiles source which I can include in the tests. Then I can run the app and set styles in the UI via webdriver, this way we could even do some visual regression testing if we wanted.

Basically for the webdriver testing I'll prefix all the 'actions' (so buttons) with `data-action="identifier"` attribute and all the inputs with `data-input="identifier"`, these shouldn't change over time as the UI changes so they should be pretty stable in the tests and not require much refactoring over time.

lukasmartinelli:
> More tests to come soon. What I want next is a tiny vector tiles source which I can include in the tests. Then I can run the app and set styles in the UI via webdriver, this way we could even do some visual regression testing if we wanted.

Easier to use a GeoJSON source is my guess.

orangemug:
> Easier to use a GeoJSON source is my guess.

Yeah that sounds a lot more sensible... :)

Issue #126, opened 2017-04-08
Add the Right-To-Left plugin


stirringhalo:
Needs the RTL plugin in order to properly display Arabic and similar languages.

orangemug:
For reference this is the plugin <https://github.com/mapbox/mapbox-gl-rtl-text>

PR #128, opened 2017-04-11
Added @mapbox/mapbox-gl-rtl-text plugin

orangemug:
`MapboxGl.setRTLTextPlugin` has a odd API where you need to load the plugin via AJAX hence the `copy-webpack-plugin` is required, fixes issue #126

**NOTE:** Adding this plugin now prevents us from running Maputnik from the `file:///` protocol, because the AJAX will fail with

```
Cross origin requests are only supported for protocol schemes: http, data, chrome, chrome-extension, https.
```

This unfortunately makes Maputnik less portable, but I don't think we really have a choice...


lukasmartinelli:
> NOTE: Adding this plugin now prevents us from running Maputnik from the file:/// protocol, because the AJAX will fail with

Hmmmm. For me this is one of the coolest feature that you can download a ZIP bundle from the releases page and just click on the index.html locally.

I know the work that went into making this possible is huge. But wonder whether it justifies breaking simple local deployment.

To make it more portable can we perhaps point the rtl-plugin to the Mapbox HTTPS url of the plugin?
If you don't have internet it will then perhaps just fail gracefully in the background.

While I really prefer having everything locally that could be a tradeoff to make so you can still just download and open the ZIP file.

lukasmartinelli:
> That's a valid point. Maybe on this one (and only one) occasion we can reach in a mess with the private API (see https://github.com/mapbox/mapbox-gl-js/blob/171db6177a6a5e7246fcef4d74ed297ee5a196d8/src/source/rtl_text_plugin.js#L31-L34)

The inspect plugin also messes with the private API :scream: :wink:

orangemug:
There you go @lukasmartinelli thats a hack for ya! No private APIs required and works on `http://` and `file:///`

src/libs/mapbox-rtl.js
```diff
@@ -0,0 +1,11 @@
+import MapboxGl from 'mapbox-gl/dist/mapbox-gl.js'
+
+// Load mapbox-gl-rtl-text using object urls without needing http://localhost for AJAX.
+const data = require("base64?mimetype=text/javascript!@mapbox/mapbox-gl-rtl-text/mapbox-gl-rtl-text.js");
```
lukasmartinelli: You smart 😎

PR #133, opened 2017-04-22
Animated loading indicator

tschaub:
Thanks for the amazing work on this project.  This pull request suggests a adding an animated loading indicator in place of the unstyled "Loading..." text.  I understand if you're not interested in adding such a thing.  In case you are, here is what it looks like (with a forced 3s load delay):

loader

### DIFF
--- a/src/template.html
+++ b/src/template.html
@@ -4,8 +4,72 @@
   <meta charset="utf-8">
   <title><%= htmlWebpackPlugin.options.title %></title>
   <meta name="viewport" content="width=device-width, initial-scale=1">
+  <style>
+    html {
+      background-color: rgb(28, 31, 36);
+    }
+    #loader,
+    #loader:before,
+    #loader:after {
+      border-radius: 50%;
+      width: 2.5em;
+      height: 2.5em;
+      -webkit-animation-fill-mode: both;
+      animation-fill-mode: both;
+      -webkit-animation: pulseload 1.8s infinite ease-in-out;
+      animation: pulseload 1.8s infinite ease-in-out;
+    }
+    #loader {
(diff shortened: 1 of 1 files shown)


orangemug:
Hey @tschaub first of all thanks for the PR. I don't think we want to force a delay because ideally we don't want to see the loading indicator at all. For example if you load it from the filesystem or in future we add a service worker it'll load instantly.

tschaub:
@orangemug - I should clarify that the change here doesn't add any additional delay.  I only added a 3 second delay for the graphic above to demonstrate the behavior when there is a delay.

orangemug:
@tschaub yeah I should have actually looked at the code, my mistake :)

The only other comment I have is that I assume the css wouldn't be linted if it's inline.

lukasmartinelli:
Thanks for contributing Tim!

orangemug:
Hey @tschaub I've fixed the peer dependency issue, if you trigger a rebuild all should now be good (same goes with #134)

orangemug:
I've restarted the CI jobs on travis for PR #133 and #134 and all now seems to be passing. I have no access to restart the jobs on appveyor, I'm assuming this is a permission issue... Does anyone ( @lukasmartinelli or @tschaub ) have the ability to restart CI jobs on appveyor? Else we might just need to push an empty commit.

orangemug:
The AppVeyor test is just failing because it's caching the node_modules. We'd need to push a new commit to make it pass. However as I've tested it locally and its fine, I think we can just merge.

Thanks @tschaub, sorry I've been away for a few months, I've just started a new job and didn't have the time.

PR #134, opened 2017-04-22
Close the "open" dialog after choosing a new style

tschaub:
This suggests closing the "open" dialog after the user chooses a new style (by uploading or from the gallery).

PR #135, opened 2017-04-28
Update react-height peer dependency version

orangemug:
This should hopefully make PRs #133 and #134 pass again. Note that https://github.com/nkbt/react-collapse/commit/24631961919e26d852120262b0cc2180e9c33ea8 seems to be the commit that caused our builds to fail.


orangemug:
Merging this PR without review as it's small and causing lots of other PR's to fail.

PR #144, opened 2017-06-23
Move babel config out of package.json

bartvde:
When using maputnik as a dependency in another project (e.g. to use the LayerEditor component), webpack doesn't read the babel config. Webpack does read it if we put it in a ```.babelrc``` file instead


lukasmartinelli:
Didn't realize that! Thanks!

orangemug:
Look good to me, merging

PR #145, opened 2017-06-23
Prevent warning for ColorField

bartvde:
This gets rid of a warning for ColorField:

```
warning.js:36 Warning: Unsupported style property background-color. Did you mean backgroundColor? Check the render method of `ColorField`.
```

### DIFF
--- a/src/components/fields/ColorField.jsx
+++ b/src/components/fields/ColorField.jsx
@@ -91,7 +91,7 @@ class ColorField extends React.Component {
     </div>
 
     var swatchStyle = {
-      "background-color": this.props.value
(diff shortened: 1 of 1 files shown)


orangemug:
@bartvde thanks for the PR

bartvde:
@orangemug thank you for the quick turnaround on these PRs

btw, are there any plans to publish maputnik on npmjs.com ?

orangemug:
Hey @bartvde, I don't think there would be any objection to publishing it to npm. The question is how would you use it? If there is a good use case please create a ticket with why you would like it published to npm and we can discuss.

PR #146, opened 2017-06-26
Add transform-class-properties plugin to babel config

bartvde:
This is a follow-up to #144 on the current code base I get (this PR fixes that):

```
ERROR in ./~/maputnik/src/components/layers/LayerEditor.jsx
Module build failed: SyntaxError: Unexpected token (45:19)

  43 | /** Layer editor supporting multiple types of layers. */
  44 | export default class LayerEditor extends React.Component {
> 45 |   static propTypes = {
     |                    ^
  46 |     layer: React.PropTypes.object.isRequired,
  47 |     sources: React.PropTypes.object,
  48 |     vectorLayers: React.PropTypes.object,
```

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,4 +1,4 @@
 {
   "presets": ["es2015", "react"],
-  "plugins": ["transform-object-rest-spread"]
+  "plugins": ["transform-object-rest-spread", "transform-class-properties"]
 }


orangemug:
Hey @bartvde thanks for another PR. If you have the time a nice edition would be to write a test for this. Something along the lines of

```
// Note: Just for demo purposes... probably incorrect. 
it("should compile via babel", function(done) {
  babel.transformFile("some-file-here.js", options, function (err, result) {
    assert.ok(err);
    done();
  });
})
```

bartvde:
good suggestion, I will see if I can get a unit test in place for this

xtuc:
I would recommend you using the https://babeljs.io/docs/plugins/transform-class-properties/ plugin rather than preset-stage-2.

bartvde:
I'm unable to find a good way to unit test this. In the current test framework it doesn't make a lot of sense since the output will get interpreted by the browser (which can't handle the import statements).
Also, I'm not even able to reproduce the issue inside of maputnik with babel-cli for instance, so it might mean that the requirement is specific to the webpack2 setup I have in that project, not sure at this point

bartvde:
@xtuc I will give that a try, thanks

bartvde:
@xtuc confirmed that also fixes my issue, will update the PR

bartvde:
PR updated, the plugin was actually already in the package.json as a dev dependency

orangemug:
@bartvde ok no worries about the testing then, thanks for giving it a try.

orangemug:
@bartvde is there anything still to do, or are we good to merge?

bartvde:
should be good to merge, thanks

PR #149, opened 2017-06-26
Make sure propTypes accept what gets passed in

### LINKED ISSUE #147: fill-translate prop gets translated into zoom field
It seems that a fill-translate prop gets translated into a zoom field somehow, I will see if I can come up with a PR to fix that, because I don't think this is intentional since it also leads to propTypes warnings

```
warning.js:36 Warning: Failed prop type: Invalid prop `value` supplied to `ZoomSpecProperty`.
    in ZoomSpecProperty (created by PropertyGroup)
    in PropertyGroup (created by LayerEditor)
    in div (created by Unknown)
    in div (created by Unknown)
    in Unknown (created by Collapse)
    in Collapse (created by UnmountClosed)
    in UnmountClosed (created by LayerEditorGroup)
    in div (created by LayerEditorGroup)
    in LayerEditorGroup (created by LayerEditor)
    in div (created by LayerEditor)
    in LayerEditor (created by Connect(LayerEditor))
    in Connect(LayerEditor)
    in div
    in Provider
```

selection_319

bartvde:
This gets rid of the propTypes warnings which you can encounter when using ```fill-antialias``` or ```fill-translate``` in a MapBox GL Style JSON document

@tbarsballe pointed out that because these props accept functions, so the current behaviour actually makes sense

closes #147

### DIFF
--- a/src/components/fields/SpecField.jsx
+++ b/src/components/fields/SpecField.jsx
@@ -43,6 +43,7 @@ export default class SpecField extends React.Component {
       React.PropTypes.string,
       React.PropTypes.number,
       React.PropTypes.array,
+      React.PropTypes.bool
     ]),
     /** Override the style of the field */
     style: React.PropTypes.object,
--- a/src/components/fields/ZoomSpecField.jsx
+++ b/src/components/fields/ZoomSpecField.jsx
@@ -31,6 +31,7 @@ export default class ZoomSpecProperty  extends React.Component {
       React.PropTypes.string,
       React.PropTypes.number,
       React.PropTypes.bool,
+      React.PropTypes.array
     ]),
   }
 


bartvde:
confirmed that both options have ```"zoom-function": true,``` in the spec doc

orangemug:
Hey @bartvde I'm a little confused by issue #147 and hence the fix could you explain a little further whats going on a how this pull request fixes the issue.

bartvde:
Sure I'll try to explain a bit more @orangemug 

If you e.g. have a ```line-translate``` field https://www.mapbox.com/mapbox-gl-js/style-spec/#paint-line-translate it supports functions, so a ZoomSpecProperty will get created. The value of this field is an array. But if you look in the propTypes: https://github.com/maputnik/editor/blob/3917a3e32392d7b134ec5c48b9178a53ee256d6d/src/components/fields/ZoomSpecField.jsx#L29:L33 it won't allow for an array right now as the value

bartvde:
A similar reasoning can be done for the other change. https://www.mapbox.com/mapbox-gl-js/style-spec/#paint-fill-antialias fill-antialias will become a SpecField which does not allow for boolean values right now in its propTypes

orangemug:
Perfect explanation, thanks @bartvde, merging.

PR #151, opened 2017-06-29
Put key on the right element

bartvde:
When you edit a dash array field you get a react warning about missing key, this PR fixes that warning


orangemug:
Thanks for the PR!

Issue #132, opened 2017-04-19
Default property


PetersonGIS:
Is the default property not available yet in Maputnik? I am not able to get it to work, though I see it implemented [here](https://jsbin.com/wagifogohi/edit?html,js,output) as part of version 0.33.0.

orangemug:
Related to #96

orangemug:
It's been added in #161

Issue #96, opened 2017-02-02
Is it possible to do data-driven styling?


chriswhong:
Maputnik is awesome, it perfectly fits our use case here at the NYC Department of City Planning where we are consuming MVTs from a self-hosted carto instance, but do not have an editor!  

Is is possible to do data-driven styles?  I am only seeing the option to pick a single color for a given layer.  Thanks, and thanks for making this tool!

muesliq:
To my knowledge it is not, yet you can always edit a layer's JSON by hand, adding those missing features by hand.

chriswhong:
Thanks!

orangemug:
This is a feature I'd also be interested in. Out of interest has anybody thought about what the UI might look like for this?

muesliq:
Probably similar to the stop functions which look like this:
![screen shot 2017-03-15 at 15 26 22](https://cloud.githubusercontent.com/assets/1606618/23953050/c64fe4ca-0993-11e7-885f-053c51d26036.png)

Just that
- A field "property" above the stop lines would be needed
- The "stops" fields wouldn't be numbers only 

So it would look something like this:

```
Color              Property: [____________]
        [DEL] [____________] [____________]
        [DEL] [____________] [____________]
                                 [Add stop]
```

What makes things really complicated is the need to be able to combine zoom functions and data driven styling, as in this example:

```
'line-offset': {
  	  'property': 'NAME',
  	  'type': 'categorical',
  	  'stops': [
  	    [{zoom: 10, value: 'LineA'}, 3],
  	    [{zoom: 10, value: 'LineB'}, 0],
  	    [{zoom: 10, value: 'LineC'}, -3],
  	    [{zoom: 14, value: 'LineA'}, 6],
  	    [{zoom: 14, value: 'LineB'}, 0],
  	    [{zoom: 14, value: 'LineC'}, -6]
  	  ]
  	}
```

I'm not sure what the ideal UI for such complexity would be. Maybe JSON is best for that, anyway.

orangemug:
@pjsier added this as a part of #161

PR #124, opened 2017-04-04
Update style spec

orangemug:
I'm a little unsure this is a 100% working PR, mainly because we're missing good test coverage, which will probably be my next task. When reviewing this PR can someone checkout the branch and give it a test locally.

PR #161, opened 2017-10-01
Add data-driven styling

pjsier:
I'm working on a project where we've been using Maputnik, and it's been great overall, but we make heavy use of data-driven styles. After looking through issue #96, I took a stab at adding that functionality here. I tried to base the design off of the conversation on that issue, and initially added the default property referenced in #132, but it looks like that may be contingent on updating the style spec.

I don't have a lot of experience with React, so feedback is more than welcome and I'm happy to make any changes


orangemug:
Hey @pjsier thanks for the pull request. I'm a little busy at the moment, but will try and review over the next couple of days. With regards to 

> ...but it looks like that may be contingent on updating the style spec.

If you're working on a project currently, could you give the branch in PR #124 a test, then we can get that merged in as well.

pjsier:
@orangemug thanks for the reply! We've actually been using the updates from that PR (along with these changes) to use the `default` property on our own instance http://eviction-lab-maputnik.s3-website-us-east-1.amazonaws.com/. We haven't run into any problems, but if there's anything in particular I can take a look at let me know

orangemug:
> We've actually been using the updates from that PR

Thats perfect thanks!

orangemug:
@pjsier I took a quick look at http://eviction-lab-maputnik.s3-website-us-east-1.amazonaws.com. First off I think it looks great, congrats this is a big step forwards for maputnik.

I think I've found a bug. I believe the background layer shouldn't support data driven styling, or at least it errors when you try to enable it `layers[0].paint.background-color: property functions not supported`

orangemug:
@chriswhong, @muesliq would you care to take a look at this PR also?

orangemug:
@pjsier FYI #124 is now merged.

pjsier:
@orangemug great to hear that branch is merged! I pulled it into this PR and re-added the `default` field. As for the bug, I hadn't been checking for the `property-function` style spec attribute to decide whether or not to show the data property button, so I added a check for that

orangemug:
Hi @pjsier sorry for taking a fews days to get back to you, I've found another issue

It looks like you're trying to set `exponential` for fields that don't support `exponential` for example  `Text layout properties -> Field` try to click the data property button and you get an error `layers[84].layout.text-field: exponential functions not supported`

Another 'nice to have' modification would be to swap the order of 'zoom function' and 'data function' buttons. A lot more fields have 'zoom function' support and looks a little cleaner in the UI.

pjsier:
@orangemug switching the order of the zoom function and data function buttons was pretty easy, so I updated that. And thanks for catching that bug with the `exponential` functions. My use case had been mostly around those, and I'd gotten tripped up on the style spec output that lists things uses `interpolated` and `piecewise-constant` instead of `exponential` and `interval`.

I updated the component to restrict the available `type` options based on the function type provided by the spec. Since either type seems to work with `categorical`, I'm defaulting to that instead of `exponential`.

I didn't include `identity` functions because when I tried that out locally it throws errors if it can't immediately parse the supplied value, and since it's based on source data validating that in advance seems like a lot of effort. If it's alright with you, I think leaving them out altogether makes sense, especially since it's the easiest one to enter into the JSON field manually.

orangemug:
Hey thanks @pjsier, that seems to be working now. Another minor issue I guess `ZoomSpecField` isn't really a good name for this file anymore as it also support data properties.

If you have the time and any ideas of a good name, a rename would be appreciated. Otherwise I think it looks good to merge. The failing tests are just issues with node v4, so I'll ignore them (I might just remove node v4 from the CI in future).

pjsier:
@orangemug great! And I was actually thinking the same thing. Based off of some of the language [Mapbox uses in the style spec](https://www.mapbox.com/mapbox-gl-js/style-spec/#types-function), maybe `FunctionSpecField`? It's not a perfect fit, but it would apply to the zoom fields as well and is a bit more general. If that works or you have any other ideas I can make that update

orangemug:
@pjsier `FunctionSpecField` is better than anything I came up with. Make the change and we'll wait for the tests to pass (except node v4) then I'll merge!

orangemug:
@pjsier ok so I reviewed this pull request again this morning and noticed another issue. Sorry my mistake I'm pretty new to this data function stuff.

The default field seems to be autocompleting to the field name. To see this turn on the data function for a field, focus the 'default field' then lose focus on that field. If you try this on 'Icon layout properties -> Size' the 'default' field' will end up being set to 'icon-size'.

Annoyingly the only way to reset the value is to edit with the JSON, but I think that is to do with our error handling generally rather than an issue with this pull request.

Note: I'll hold off merging #167 until this is merged, so you don't have to deal with the conflicts.

pjsier:
@orangemug ah thanks for catching that. It looks like it was mostly a typo (I was only handling one function parameter but `SpecField` was returning 2), but I also added a check in to `changeDataProperty` to allow people to empty the color field

src/components/fields/FunctionSpecField.jsx
```diff
@@ -0,0 +1,358 @@
...
+      delete this.props.value[propName]
+    }
+    this.props.onChange(this.props.fieldName, this.props.value)
+  }
+
+  getDataInput(value, dataLevel, zoomLevel) {
+    const dataProps = {
+      label: "Data value",
+      value: dataLevel,
+      onChange: newData => this.changeStop(idx, { zoom: zoomLevel, value: newData }, value)
```
orangemug: @pjsier I think you mean `zoomLevel` here.

orangemug:
@pjsier I've added a comment, in the code. Also noticed that some of the properties are throwing errors, updating `mapbox-gl^0.40.1` and `@mapbox/mapbox-gl-style-spec^9.0.1` fixes those I believe.

We've now got a new error, modifying data function values causes the error `array expected, undefined found`

src/components/fields/FunctionSpecField.jsx
pjsier: @orangemug actually, I think I was accidentally doing some fancy scoping, because it does need to access the `idx` of the parent to change the stop. About to update that by removing the `getDataInput` and moving that conditional into the loop so it's more explicit

pjsier:
@orangemug I updated `mapbox-gl` and the style spec and clarified that scope, but I'm not seeing the `array expected` error that you mentioned. Could you give an example of a layer it's happening on?

orangemug:
@pjsier I can't seem to reproduce the issue now. Ok this looks great, there are a couple of UI things I think we could improve

 - Labels on the 3 text boxes, it's not clear what to enter where at the moment
 - If you forget to enter a property it'll error in odd ways. Maybe we should disable the stops until the property is selected.
 - Some UI tests. We need to discuss testing in general the project is pretty untested at the moment 

However we can pick those things up later on in new tickets.

Congrats on the feature, this is a big step forward!

orangemug:
> I'm working on a project where we've been using Maputnik, and it's been great overall, but we make heavy use of data-driven styles.

Oh also @pjsier it'd be great to hear about what you're working on over in issue #164

PR #165, opened 2017-10-08
Modal scrolling #156

gregorywolanski:
Help wanted (label : ), help provided. 

Please let me know if there is anything wrong with this pull request. I’m new to open source.

PR #166, opened 2017-10-10
Added 'Load from URL' option in open modal

### LINKED ISSUE #120: Need an easy way to share styles
Whenever I work with styles, I constantly need to share my style with others.  While it would be awesome to have git integration, pull requests, etc, available, the very basic usecase is to be able to send someone my work in progress, so that they can take a look at a specific issue / modify data source / recommend style change / file a bug with a permanent link to a style.

How can we make it so that uploading to GIST (for example) can be auto-loadable, without requiring the user to save the gist to a file and upload the file to the same editor?

orangemug:
This add a 'Load from URL' option to the open modal as requested in #120.

Note this makes the modal content larger so we're waiting on #165 to be merged first.


pjsier:
Would it make sense to add a "Share this Style" button after you've saved a style to an anonymous gist? It could show the URL with the `style` parameter pre-populated, and copy it to the clipboard if you click a button just to make it more apparent that people can do that

orangemug:
@pjsier yeah that makes sense to me, could you raise an issue.

PR #167, opened 2017-10-10
Fix zoom field order issue

### LINKED ISSUE #116: Unable to enter zoom field
if my first zoom is set to 2, it is not possible to enter a second zoom above 10, because typing 1 automatically treats it as an error (it is lower than 2), and resets it to 10. I think the field should not reset on error, but instead show a red border/background, and wait for the user to leave the field. Once left, then you can reset the value.  Also, invalid field should be excluded from automatic refresh until it becomes valid again (either after leave+reset, or once the value is valid again)

orangemug:
This should fix #116. When editing zoom stops it now re-orders in the UI and focus is maintained on the field being edited.

### DIFF
--- a/src/components/fields/ZoomSpecField.jsx
+++ b/src/components/fields/ZoomSpecField.jsx
@@ -13,6 +13,11 @@ import FunctionIcon from 'react-icons/lib/md/functions'
 
 import capitalize from 'lodash.capitalize'
 
+let REF = 0;
+function getUniqueRef() {
+  return "r"+REF++;
+}
+
 function isZoomField(value) {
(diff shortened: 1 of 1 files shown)


orangemug:
Waiting for #161 to be merged first to prevent conflicts in that PR.

orangemug:
Closing this pull request as it isn't the right approach. We need to persist a key for react somewhere in the JSON data. I'm not entirely sure the best method to do that at the moment. Relying on `zoom` isn't enough because during editing it's possible for the same zoom level to appear twice in the `stops`list.

PR #172, opened 2017-10-12
Fix insecure asset loading

### LINKED ISSUE #153: Mapzen tiles URL incorrect
When adding the Mapzen tiles as a data source, it proposes this URL:

`http://tile.mapzen.com/mapzen/vector/v1/{layers}/{z}/{x}/{y}.pbf?api_key=mapzen-RVcyVL7`

Two issues:

1. `{layers}` should be replaced, probably by `all`
2. `.pbf` should be `.mvt`

Hence:

http://tile.mapzen.com/mapzen/vector/v1/all/{z}/{x}/{y}.mvt?api_key=mapzen-RVcyVL7

pjsier:
Changes the HTTP paths for loading thumbnail images and one tile layer to HTTPS to avoid Mixed Content errors. Also closes #153, but it looks like there's a separate issue with the Tilezen style where it returns HTTP 401 unauthorized errors for the Mapbox fonts, so I'm guessing the API key for that style may be invalid here https://rawgit.com/lukasmartinelli/tilezen-gl-style/master/style.json

### DIFF
--- a/src/config/styles.json
+++ b/src/config/styles.json
@@ -3,25 +3,25 @@
     "id": "klokantech-basic",
     "title": "Klokantech Basic",
     "url": "https://rawgit.com/openmaptiles/klokantech-basic-gl-style/master/style.json",
-    "thumbnail": "http://maputnik.com/thumbnails/klokantech-basic.png"
+    "thumbnail": "https://maputnik.github.io/thumbnails/klokantech-basic.png"
   },
   {
     "id": "dark-matter",
     "title": "Dark Matter",
     "url": "https://rawgit.com/openmaptiles/dark-matter-gl-style/master/style.json",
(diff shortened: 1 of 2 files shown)


orangemug:
> Tilezen style where it returns HTTP 401 unauthorized errors

@pjsier yeah I took a quick look at that a while back. I'm not sure where that api key came from. Maybe it's better to comment out Tilezen for the time being as it doesn't work... thoughts?

pjsier:
@orangemug that makes sense to me. Should I remove the Tilezen tileset and style?

orangemug:
> Should I remove the Tilezen tileset and style?

@pjsier yeah I think lets remove it and we'll leave #153 open if someone needs it there's an issue there with all the info

pjsier:
@orangemug makes sense! Just removed them

orangemug:
Thanks, merged

PR #174, opened 2017-10-16
Add share style link, copy button

### LINKED ISSUE #171: Add "Share this Style" button with pre-populated URL including style
Building off of the functionality requested in #120, add a "Share this Style" button after you've saved a style to an anonymous gist. It could show the URL with the style parameter pre-populated, and copy it to the clipboard if you click a button to make it more apparent that the functionality exists

pjsier:
Closes #171, and adds to functionality mentioned in #120. Adds a copy to clipboard button and an input with the URL including a pre-populated style parameter after a user exports a style to a gist

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
     "react-codemirror": "^0.3.0",
     "react-collapse": "^4.0.3",
     "react-color": "^2.10.0",
+    "react-copy-to-clipboard": "^5.0.1",
     "react-dom": "^15.4.0",
     "react-file-reader-input": "^1.1.0",
     "react-height": "^3.0.0",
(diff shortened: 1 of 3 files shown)


orangemug:
It's not strictly necessary here but it might be worth adding `rel="noopener noreferrer"` to the link for future robustness (see <https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/>)

pjsier:
That sounds like a good idea to me, but should we create a separate issue/PR for it? I could go either way since it looks like all of the `target="_blank"` links are in ExportModal.jsx, and it doesn't seem to have any side effects

orangemug:
@pjsier yeah you're right, it's else where too (`src/components/Toolbar.jsx`). I'll merge an fix it in another PR

PR #176, opened 2017-10-16
Added rel="noopener noreferrer" to external links.

orangemug:
See <https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/> for details

### DIFF
--- a/src/components/Toolbar.jsx
+++ b/src/components/Toolbar.jsx
@@ -32,7 +32,8 @@ function ToolbarLink(props) {
   return <a
     className={classnames('maputnik-toolbar-link', props.className)}
     href={props.href}
-    target={"blank"}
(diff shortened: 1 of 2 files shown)


src/components/Toolbar.jsx
```diff
@@ -32,7 +32,8 @@ function ToolbarLink(props) {
   return <a
     className={classnames('maputnik-toolbar-link', props.className)}
     href={props.href}
-    target={"blank"}
+    rel="noopener noreferrer"
+    target={"_blank"}
```
pjsier: Can this just be `target="_blank"` since it's not pulling from `props`? I tried that locally and it works

orangemug: Yup, fixed :)

Issue #87, opened 2017-01-17
Display zoomlevel next to attribution


rroset:
It would be useful to have the current zoomlevel displayed nexto to attribution so that one could see what ZL the current view belongs to.
It would help when dealing with ramps and finetuning the entry and exit ZL for certain layers
Could be something like
ZL: xx.x

lukasmartinelli:
Absolutely. Always looked at the URL hash now but that would be a good additional control.

gregorywolanski:
@rroset Do you have in mind something similar to the thing you can see on the attached screenshot?

![1](https://user-images.githubusercontent.com/787518/31737914-459d7f52-b449-11e7-92a3-8702abe26b1b.png)

rroset:
Sure, that would do.
Another option would be to add it as text in the attribution line, like in here:
http://betaserver.icgc.cat/vectortiles/mtc25m14_elevation.html#10.99/41.5972/2.364

I'm happy with any of those =)
thanks

orangemug:
@gregorywolanski submitted a PR which has been merged. Closing the issue.

rroset:
thanks you all!

gregorywolanski:
BTW @rroset Silly question, but I’m not a pro Maputnik user myself and I’m curious: What do you need the X.X (float) accuracy of zoom level for? Why not an integer?

Also, I just realized that I implemented X.XX zoom level accuracy instead of X.X (or "XX.X", as you put it). I can change it.

tomass:
@gregorywolanski one of advantages of vector tiles is that you are not limited to fixed zoom levels. You can draw a map at zoom level 10.12 and it would be at different scale as compared to 10.23.

If you want to display two points and a route between them on the map, you can calculate exact zoom level (which would usually be float), so that both points fit in a viewport and both of them are at the edge of the viewport thus making all objects as large as possible for better visualisation.

PR #179, opened 2017-10-19
Display zoom level #87

gregorywolanski:
Response to #87:


orangemug:
Apologies it took me a while to get to this, it looks great. Merging.

gregorywolanski:
@orangemug There is no need to apologize. :) One is glad to be of service. :D

Issue #130, opened 2017-04-12
expand/collapse all layer groups


pathmapper:
It would be useful to be able to expand/collapse all layer groups with one click or keyboard shortcut. 

Current situation example: 
If you want to work on all symbol layers within a style, you have to expand all layer groups manually not to miss one of them.

gregorywolanski:
I don't know where the icons used in Maputnik come from (@lukasmartinelli? @orangemug?) so I allowed myself to use text label.

What do you think about it working and looking like on the attached GIF?

I’m open to feedback before creating a pull request. :)

![2017-10-24 20_45_34](https://user-images.githubusercontent.com/787518/31961927-7f10b83c-b8fc-11e7-9059-8943812eeb96.gif)

pathmapper:
@gregorywolanski wow, very nice! 

I like the place where you put the icon and I think your text label (Expand/Collapse) fits very well with the "Add Layer" button.

Thanks a lot!

orangemug:
Merged @gregorywolanski PR #185 released in `v1.1.0-beta2` https://maputnik.github.io/releases/v1.1.0-beta2

Note: I'll release v1.1.0 properly (to https://maputnik.github.io/editor) once I've given everybody a change to raise any further issues.

PR #185, opened 2017-10-25
Add expand/collapse all layer groups feature (#130)

gregorywolanski:
Response to #130:

31961927-7f10b83c-b8fc-11e7-9059-8943812eeb96

### DIFF
--- a/src/components/layers/LayerList.jsx
+++ b/src/components/layers/LayerList.jsx
@@ -49,6 +49,7 @@ class LayerListContainer extends React.Component {
     super(props)
     this.state = {
       collapsedGroups: {},
+      areAllGroupsExpanded: false,
       isOpen: {
         add: false,
       }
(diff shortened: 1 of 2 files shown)


orangemug:
First off sorry it took me so long to review this pull request, I was planning to change the UI a little and wanted to make sure it wouldn't conflict (it doesn't).

The functionality of the expand/collapse looks good. Although I have a couple of comments about the UI.

The button is a toggle button. Toggle buttons usually represent the state, but here we actually have 3 states rather than 2.

 - All expanded
 - All collapsed
 - Some expanded / some collapsed (the missing one)

The button also feels like something you wouldn't use often, so should it be such a big part of the main UI?

An alternative would be to put a little menu kind of like the one shown here <https://material.io/guidelines/components/menus.html#menus-usage> with 2 options

 - Expand all
 - Collapse all

What do you think @gregorywolanski does this make sense and is it worth the extra effort?

orangemug:
Thanks for the detailed reply and I think it makes sense. My reason for wanting a context menu was additionally that I think there may be more options later down the road... but that's not the problem at hand so I probably shouldn't try to predict the future.

Lets stick with your solution.

Although there is a styling issue, expand should sit to the right with the layers button (as in your screenshot)

screen shot 2017-11-29 at 21 47 46

gregorywolanski:
@orangemug I’ll fix it. What browser do you use?

orangemug:
> What browser do you use, on which OS?
Latest chrome/firefox on the mac

gregorywolanski:
Thanks, @orangemug. Silly mistake. Fixed. :)

gregorywolanski:
Happy New Year, @orangemug! :)

I’m looking forward to reading about the changes in the UI you have planned.

Issue #95, opened 2017-02-02
Inspect Mode doesn't work after style change


muesliq:
Steps to reproduce bug:

1. Load Maputnik with Mapbox based style
1. Click "Open" and select Openmaptiles based style
1. Click "Inspect Mode"

Console error message:
`Error: layers: source "mapbox" not found `

(On a side note maybe the other menu buttons should be greyed out when in Inspect Mode.)

gregorywolanski:
I’m working on it.

orangemug:
@gregorywolanski fixed this in #222

PR #186, opened 2017-10-27
Inspect Mode doesn't work after style change (#95) bug fix

gregorywolanski:
Help wanted, help provided. :)

### DIFF
--- a/src/components/map/MapboxGlMap.jsx
+++ b/src/components/map/MapboxGlMap.jsx
@@ -112,7 +112,23 @@ export default class MapboxGlMap extends React.Component {
     const nav = new MapboxGl.NavigationControl();
     map.addControl(nav, 'top-right');
 
-    const inspect = new MapboxInspect({
+    map.on("style.load", () => {
+      const inspect = this.generateInspectControl()
+      map.addControl(inspect)
+
+      this.setState({ map, inspect })
+    })
+
+    map.on("data", e => {
+      if(e.dataType !== 'tile') return
+      this.props.onDataChange({
+        map: this.state.map
+      })
+    })
+  }
+
+  generateInspectControl = () => {
+    return new MapboxInspect({
       popup: new MapboxGl.Popup({
         closeOnClick: false
       }),
@@ -131,18 +147,6 @@ export default class MapboxGlMap extends React.Component {
           return renderLayerPopup(features)
         }
       }
-    })
-    map.addControl(inspect)
-
-    map.on("style.load", () => {
-      this.setState({ map, inspect });
-    })
-
-    map.on("data", e => {
-      if(e.dataType !== 'tile') return
-      this.props.onDataChange({
-        map: this.state.map
-      })
     })
   }
 
--- a/src/components/map/MapboxGlMap.jsx
+++ b/src/components/map/MapboxGlMap.jsx
@@ -113,6 +113,10 @@ export default class MapboxGlMap extends React.Component {
     map.addControl(nav, 'top-right');
 
     map.on("style.load", () => {
+      if(this.state.inspect) {
+        map.removeControl(this.state.inspect)
+      }
+
       const inspect = this.generateInspectControl()
       map.addControl(inspect)
 


src/components/map/MapboxGlMap.jsx
```diff
@@ -112,7 +112,23 @@ export default class MapboxGlMap extends React.Component {
     const nav = new MapboxGl.NavigationControl();
     map.addControl(nav, 'top-right');
 
-    const inspect = new MapboxInspect({
+    map.on("style.load", () => {
+      const inspect = this.generateInspectControl()
+      map.addControl(inspect)
```
orangemug: @gregorywolanski some questions

 - can you explain why this works, I'm a little confused why this fixes the issue
 - do we also want to be running `map.removeControl` at some point.

gregorywolanski: ### Why this works?

English is not my primary language so don't hesitate to ask clarifying questions. :)

Bear with me: mapbox-gl-inspect (MapboxInspect) uses map styles. Map styles contain sources ("openstreetmap", "mapbox"). Right now, without the bug fix, Maputnik starts with one map style and invoke MapboxInspect only once. It works OK as long as you use map styles that use the same source. In this scenario, when you change map style, the source list remains the same. However, when you switch to a map style with a different source (as in the example provided by @muesliq), MapboxInspect is not updated, uses the old list of sources, which lacks the new source and the app returns `Error: layers: source "mapbox" not found`. The bug fix enables MapboxInspect to be updated every time the user change the style.

* * *

### Do we also want to be running `map.removeControl` at some point?

Good question. I haven't thought about that, because of no organoleptic signs that it can lead to problems.  I'll add today `map.removeControl` before `map.addControl` if `this.inspect` has been defined.

orangemug: > Right now, without the bug fix, Maputnik starts with one map style and invoke MapboxInspect only once. It works OK as long as you use map styles that use the same source.

Thanks for the context. So I think adding / removing sources from the current style also causes the same issue. You can see this if you follow the steps

 1. Load the empty map style
 2. Add openmaptiles source
 3. Go into inspect mode
 4. Remove openmaptiles source
 5. Go into inspect mode again

You'll see the source that no longer exists is still being inspected.

I think the problem actually lies in [lukasmartinelli/mapbox-gl-inspect](https://github.com/lukasmartinelli/mapbox-gl-inspect) because this does the tracking of changes to the sources list <https://github.com/lukasmartinelli/mapbox-gl-inspect/blob/master/lib/MapboxInspect.js#L203-L204>.

I think the real issue might be here <https://github.com/lukasmartinelli/mapbox-gl-inspect/blob/master/lib/MapboxInspect.js#L128-L148> but I'm unsure

Thoughts?

orangemug: _

orangemug:
@gregorywolanski note you'll have to expand click 'Show outdated' to see my comment

It's an odd github 'feature' new comments to old threads aren't shown in the main conversation by default

src/components/map/MapboxGlMap.jsx
gregorywolanski: Yes, you’re right, adding and removing sources from the current style also causes the same issue.

First observation: https://github.com/lukasmartinelli/mapbox-gl-inspect/blob/master/lib/MapboxInspect.js#L128-L148 looks like it’s about a different type of sources list. I mean: The subject of the bug is an array that looks like `["mapbox", "openstreetmap"]`. The piece of code that you mentioned operates on objects like `{openmaptiles: Array(15)}`. 

I’m digging into [lukasmartinelli/mapbox-gl-inspect](https://github.com/lukasmartinelli/mapbox-gl-inspect).

orangemug: > looks like it’s about a different type of sources list.

Yeah sorry having a quick look and pointing at random code probably wasn't that helpful :)

> I’m digging into lukasmartinelli/mapbox-gl-inspect.

 Yay! Just shout if you want me to test anything

gregorywolanski: You were right, it’s a problem that can be solved inside [lukasmartinelli/mapbox-gl-inspect](https://github.com/lukasmartinelli/mapbox-gl-inspect).

It took me an embarrassing number of hours to debug, but I think that I found the problem and the solution.

I’ll get some sleep, test it once again, prepare a pull request in [lukasmartinelli/mapbox-gl-inspect](https://github.com/lukasmartinelli/mapbox-gl-inspect) and then close this pull request.

It’s a pleasure to work with you, @orangemug. :)

orangemug: > ...but I think that I found the problem and the solution

Yay, congrats @gregorywolanski! If you'd like me to give it a quick test just mention me in the new PR

gregorywolanski: It turned out my solution doesn’t remove the empty map style problem. :/ :)

I’ll ask you for a quick test in a few days, thank you.

PR #189, opened 2017-11-04
Disable polling by default

orangemug:
I've disabled `webpack-dev-server` polling by default. Most users won't need this enabled and for those that don't require polling it was causing lots of CPU usage. Which was causing my cafe trips to work on maputnik to be cut short due to a drained battery  ☹️

This should be a better default for most users. I've also verified works inside a docker container just fine.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,12 @@ npm install
 npm start
 ```
 
-Build a production package for distribution.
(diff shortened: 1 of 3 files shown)


orangemug:
The tests are failing in nodejs v4. This isn't an issue with this PR so merging.

PR #191, opened 2017-11-06
Fixed startup errors/warnings

orangemug:
I've fixed all the warnings / errors on initial load. Here's what the PR includes, its a little confusing so the links to the commits are also linked inline.

> Warning: Failed prop type: The prop `children` is marked as required in `LayerEditorGroup`, but its value is `undefined`.

Fixed field name (it was incorrectly named LayerEditorGroup) and removed `children` as a required field.
 <https://github.com/orangemug/editor/commit/e148607c7a11e68a4c72981b8d48a08363cee1db>

> Warning: Failed prop type: The prop `value` is marked as required in `MinZoomBlock`, but its value is `undefined`.
> Warning: Failed prop type: The prop `value` is marked as required in `MaxZoomBlock`, but its value is `undefined`.

The layer minzoom/maxzoom has no default therefore can be undefined. Removed as a required prop.
<https://github.com/orangemug/editor/commit/bb4f3482ad299ea9a996b4f0e6c1dfadfb21190b>

> ./\~/mapbox-gl/dist/mapbox-gl.js. This seems to be a pre-built javascript file. Though this is possible, it's not recommended. Try to require the original source to get better results.
> client:68 ./~/openlayers/dist/ol.js. This seems to be a pre-built javascript file. Though this is possible, it's not recommended. Try to require the original source to get better results.

Added `noParse` to webpack config for prebuilt files. <https://github.com/orangemug/editor/commit/9801f49f4ec5d82432ff34692d5a4b0c72cf090c>

> Color field has no element to adjust position

This is normal usage of the component (<https://github.com/maputnik/editor/blob/master/src/components/fields/ColorField.jsx#L40>), so I removed the warning. <https://github.com/orangemug/editor/commit/365a0518a545ec0dbd0d470a377be8c90c2ef422>

> Warning: Accessing PropTypes via the main React package is deprecated, and will be removed in  React v16.0. <https://github.com/orangemug/editor/commit/9bc603a5106ddb5a2fbef0f9bbaebb0f877c24d1>

Bumped `react-codemirror@1.0.0` and `react-sortable-hoc@0.6.8`. Also updated code in all the components. <https://github.com/orangemug/editor/commit/a4c6a1835393f45fe7e8f858d58106c3d146fb08>

### DIFF
--- a/config/webpack.config.js
+++ b/config/webpack.config.js
@@ -23,7 +23,11 @@ module.exports = {
     extensions: ['', '.js', '.jsx']
   },
   module: {
-    loaders
+    noParse: [
+      /mapbox-gl\/dist\/mapbox-gl.js/,
+      /openlayers\/dist\/ol.js/
+    ],
+    loaders: loaders
   },
   node: {
     fs: "empty",
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
     "mousetrap": "^1.6.0",
     "ol-mapbox-style": "1.0.1",
     "openlayers": "^3.19.1",
+    "prop-types": "^15.6.0",
     "react": "^15.4.0",
     "react-addons-pure-render-mixin": "^15.4.0",
     "react-autocomplete": "^1.4.0",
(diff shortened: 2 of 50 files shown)


orangemug:
The tests are failing in nodejs v4. This isn't an issue with this PR so merging.

PR #193, opened 2017-11-08
Disable pointer-events on doc tooltips

orangemug:
Currently large documentation tooltips overlap the labels below them. This changes the behaviour so that the mouse ignores the tooltip. This allows the user to easier pan over the labels and see the docs.


orangemug:
See <https://42-84182601-gh.circle-artifacts.com/0/tmp/circle-artifacts.wCO1FrC/build/index.html#12.24/47.3718/8.5481> for demo

PR #200, opened 2017-11-29
Fix zoom field input ordering (#116)

### LINKED ISSUE #116: Unable to enter zoom field

orangemug:
This pull request orders zoom field inputs during edit. An example, if you have the following zoom stops

```
2: #fff
10: #000   <= Keyboard focus
```

Then you delete the `0` from the `10` the zoom fields will be reordered keeping the focus on the stop you were just editing.

```
1: #000   <= Keyboard focus
2: #fff
```

The biggest change is that I've also split the data/zoom properties into their own files.

You can see the build here <https://111-84182601-gh.circle-artifacts.com/0/tmp/circle-artifacts.rbyROMd/build/index.html>

Fixes #116


orangemug:
Does anybody have any feedback @pjsier @nyurik ?

You can see the build at <https://111-84182601-gh.circle-artifacts.com/0/tmp/circle-artifacts.rbyROMd/build/index.html>

orangemug:
Ooops I meant to add a comment saying

> Don't merge just yet!

Nevermind @pjsier any chance you could to test the data property functionality as you originally submitted that PR. It seems to work :/

Build here <https://111-84182601-gh.circle-artifacts.com/0/tmp/circle-artifacts.rbyROMd/build/index.html>

FYI: I'm in the process writing automated tests so I can be more confident with merges.

pjsier:
@orangemug I should be able to check that in the next few days! Also, it may actually be worth changing the data field to a JSON editor now that [expressions are part of the style spec](https://www.mapbox.com/mapbox-gl-js/style-spec/#expressions) and it looks like the existing data property functions are going to be deprecated

pjsier:
@orangemug doesn't look like the functionality is working the way you describe it above for changing the zoom level. I tried changing it for a function spec field (from 10 to 1), and I got the standard error about zoom levels needing to be in order

nyurik:
@pjsier could you describe the steps?  I tried it and it seems to be working for me.  E.g. background paint property, switch to per-zoom levels, and add one extra level (e.g. 2 and 10), and changing the zoom from 10 to 1 -- it reorders.

pjsier:
@nyurik it works for me with the zoom spec field, but if you click the function spec field button instead, select a data attribute, add an extra level, and then change from 10 to 1 it throws an error.

nyurik:
@pjsier I must be missing something.  This is where I tried it:
image

pjsier:
@nyurik Ah yeah that's the zoom field, but the function/data property field where it's failing looks like this:

screen shot 2017-11-30 at 5 46 11 pm

nyurik:
@pjsier, thanks, my apologies, I was looking at the background type only, forgot about the other ones. I guess it should be a separate patch, as the behavior is as it was before. Will keep the issue open.

Issue #60, opened 2017-01-10
Layer tooltips should be clickable


muesliq:
It would be a big improvement if layers were clickable, opening the corresponding layer in the sidebar. 

![screen shot 2017-01-10 at 14 19 45](https://cloud.githubusercontent.com/assets/1606618/21807671/38f5a748-d740-11e6-82cc-5ad14578a417.png)

(Better even: Highlight the layer in the sidebar already on hover.)

klokan:
+1

gregorywolanski:
I'm working on it. I need to get https://github.com/lukasmartinelli/mapbox-gl-inspect/pull/8 merged to finish this enhancement. :)

orangemug:
Released in [v1.1.0-beta2](https://maputnik.github.io/releases/v1.1.0-beta2)

PR #213, opened 2018-01-08
Clickable layer tooltips (#60)

gregorywolanski:
The not obvious piece of code is responsible for ensuring that when you click on a layer that is a part of a collapsed layer group, the layer group will expand on click as well.


orangemug:
@gregorywolanski this works great! Another PR has caused a conflict with `package.json`, fix that and we're good to merge.

orangemug:
Released in `v1.1.0-beta2` see https://maputnik.github.io/releases/v1.1.0-beta2

Issue #212, opened 2018-01-04
Preview Export to bl.ocks.org not working


keum:
I'm trying to Preview stylized map using Export feature with Tilehosting https://www.tilehosting.com/ API and got it to work few weeks ago but now getting blank screen at bl.ocks.org site now. Anyone having similar issue?

keum:
Currently using Export style option save to Gist preview link shows blank map as stated above, in our testing to downloaded file to display locally, we were able to show the basemap by changing 
https://api.mapbox.com/mapbox-gl-js/v0.28.0/mapbox-gl.js
https://api.mapbox.com/mapbox-gl-js/v0.28.0/mapbox-gl.css
v0.28.0 -> v0.43.0 
we would like this to be updated so upcoming Maputnik demonstration, people can display their map on-line using Export to preview function. 
Please help.

orangemug:
> upcoming Maputnik demonstration
@keum I'll try find some time to sort it out. When is your demonstration?

keum:
thanks @orangemug in two weeks. One solution is to download locally and change those api version to 0.43.0 and spin local web machine and open the map that way but...it's so much nicer w/ bl.ocks.org site and easier to share w/ other participants. thanks!

keum:
@orangemug here is screen shots to tell what i'm describing in case it's not too clear. 

![image](https://user-images.githubusercontent.com/1133652/34616664-c65f34be-f1ed-11e7-9fc5-ebf734fcfa1f.png)

keum:
@orangemug oops..forgot to add second image
![image](https://user-images.githubusercontent.com/1133652/34616734-f3156f6e-f1ed-11e7-8737-88b0a4bc7702.png)

thanks for looking into it. really appreciate it.

orangemug:
Hey @keum I've made a beta v1.1.0 release with that changes you asked for, it is available at https://maputnik.github.io/editor-v1.1.0-beta I'll let some people give it a test before I release to `/editor` feel free to use <https://maputnik.github.io/editor-v1.1.0-beta> for your demo.

Sorry it took so long, I've been a little under the weather and haven't found any time. Hopefully I've got these to you just in time, for your demo.

pkgeo:
hi @orangemug, thank you much! Yes, it works to export into blocks.org to preview fine. But i found another issue in your beta version.  In add layer dialog box, it appears that you can not select the source nor able to type in the source (only first letter shows up).  Thanks for all your work, really appreciate it!
[...]

PR #214, opened 2018-01-12
Fix for #212 - outdated GL JS rev in export breaks bl.ocks preview

tobinbradley:
Fix for #212 broken bl.ocks preview. Note I upped it to current - 0.43.0. Should it be 0.41.0 to be consistent with current maputnik? I don't know if there are any breaking changes effecting maputnik between 0.41 and 0.43.

### DIFF
--- a/src/components/modals/ExportModal.jsx
+++ b/src/components/modals/ExportModal.jsx
@@ -56,8 +56,8 @@ class Gist extends React.Component {
     <meta charset="UTF-8">
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <title>`+styleTitle+` Preview</title>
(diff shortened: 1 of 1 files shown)


orangemug:
@tobinbradley I've added your commit to pull request #217. It'll get merged with that pull request.

Thanks for contributing!

orangemug:
@tobinbradley another note, the branch you're making the pull request from. It's better if you create your own branch rather than creating a pull request from tobinbradley:master. This way you can create multiple pull requests to maputnik/editor. There is a guide here https://gist.github.com/Chaser324/ce0505fbed06b947d962

Anyway that's one for next time, if you contribute again, which would be nice 😊. Thanks for your contribution.

tobinbradley:
No worries, and thanks for the tip!

Issue #229, opened 2018-01-30
Adding a style without Glyphs defined throws an exception with no feedback in interface


justenPalmer:
To replicate, upload a style.json file without a property "glyph". Nothing happens in the interface and console logs the following exception:

Uncaught TypeError: Cannot read property 'replace' of undefined

pathmapper:
@justenPalmer just to be sure: 
Does your style contains a layer which uses the `text-field` layout property?

From the [style-spec](https://www.mapbox.com/mapbox-gl-js/style-spec/#root-glyphs):

> This property is required if any layer uses the `text-field` layout property.

justenPalmer:
Actually the style.json file I am using only has one layer defined which is the `background` layer. It has no `text-field` thus should not require the `glyph` property according to the spec.

orangemug:
Fixed by @justenPalmer in #231

PR #231, opened 2018-01-31
Issue 229: Adding a style without Glyphs defined throws an exception with no feedback in interface

justenPalmer:


justenPalmer:
It looks like the travis-ci build failed due to a timeout. Could someone queue this to try the build again?

orangemug:
Thanks for the pull request @justenPalmer, much appreciated.

Question is a missing `glyphs` in the root on the style actually an error? From the spec it looks it's optional (also you mention this in #229). Should the code not just handle a style without a glyphs key present. 

Let me know your thoughts?

pathmapper:
> Should the code not just handle a style without a glyphs key present.

This is also how I understand the spec. 

Here is a style to reproduce the error in Maputnik [style_only_bg.zip](https://github.com/maputnik/editor/files/1684736/style_only_bg.zip)

Uploading this style to Mapbox Studio works fine without errors.

justenPalmer:
@orangemug, yeah I do agree. I'll try to fix the underlying exception that gets thrown. Maybe we should also do a check to see if any layers use text and display an error if there is no glyphs then.

orangemug:
> I'll try to fix the underlying exception that gets thrown

Sounds good thanks @justenPalmer for taking a look.

> Maybe we should also do a check to see if any layers use text and display an error if there is no glyphs then.

Hmm, so I would imagine that either the [@mapbox/mapbox-gl-style-spec](https://www.npmjs.com/package/@mapbox/mapbox-gl-style-spec) validation would do this for us. Or mapbox-gl-js would throw an error. Feel free to take a look

justenPalmer:
Ok, I've updated the pull request to reflect the discussion in this thread. Let me know if you guys would like any further changes.

src/components/App.jsx
```diff
@@ -112,23 +114,16 @@ export default class App extends React.Component {
...
-      this.updateFonts(newStyle.glyphs)
-    }
-    if(newStyle.sprite !== this.state.mapStyle.sprite) {
-      this.updateIcons(newStyle.sprite)
-    }
-
     const errors = styleSpec.validate(newStyle, styleSpec.latest)
     if(errors.length === 0) {
+
+      if(newStyle.glyphs !== this.state.mapStyle.glyphs) {
```
orangemug: Hmmm, I wonder if it's maybe it's safer to just do

```
if(newStyle.hasOwnProperty("glyphs") && newStyle.glyphs !== this.state.mapStyle.glyphs) { ... }
```

...and leave it above the `(errors.length === 0)` check. My thinking is that if the style has errors does mapbox-gl-js always fail to render, or does it sometimes do a best effort to render the style?

Thoughts @justenPalmer?

justenPalmer: I thought about that, but then the app state would not get updated if the glyphs property was absent on the new style. The app would still think the glyphs property from the previous style was the active one. I think it's better just to update the state with the undefined property if the glyphs is not there, but I'm not sure the impact an undefined value could have on the interface (when testing, I didn't notice any issues).

As for moving the update glyphs and update sprite above the style validation, it seems like it would be strange to update glyphs and then have the new style not pass validation. Wouldn't that also create a state where the glyphs and sprite properties could be out of sync with the active style?

orangemug:
Looks good to me, AppVeyor is just playing up so we'll ignore it. Thanks for the contribution

src/components/App.jsx
orangemug: That all makes sense, and seems well thought through. Thanks for the explanation.

Issue #164, opened 2017-10-05
Show & tell


orangemug:
Hi Maputnik community,

I think it'd be great if we could share the work being created by all the talented people using Maputnik.

If you're a company, organisation or individual please share your work. What I'm looking for

 - Images / links to maps made with Maputnik
 - Testimonials - Why is Maputnik useful for you and your team?

It'd also be great to get a list of companies and organisations using Maputnik. Later we can publish this to the homepage.

So please share all your creations in the comments below.

orangemug:
@pjsier @chriswhong @bartvde I thought you all might want to share your work. If you're not interested, feel free to ignore

bartvde:
@orangemug we have actually taken a different approach since and are not using maputnik anymore

orangemug:
@bartvde ok, thanks for the reply.

pjsier:
@orangemug I work with an organization called Eviction Lab on an open source web app visualizing eviction data across the United States (more information here https://preview.evictionlab.org/). It's currently in development ([GitHub org here](https://github.com/EvictionLab)), but Maputnik has been great for making it easier for designers and developers to collaborate on map styles. It's been especially useful for us because we're using a lot of data driven styles and custom tilesets

Side note on that: when will the recent PRs be deployed to maputnik.github.io?

orangemug:
Thanks @pjsier.

> Side note on that: when will the recent PRs be deployed to maputnik.github.io?

Yeah apologies I've been a bit slow on that I'll try to get on to that tomorrow. On yet another related note @lukasmartinelli can you make @pjsier a member of the maputnik org

pjsier:
@orangemug no problem! Is there an issue out there for automating deployments with Travis? And, out of curiosity, is editor.openmaptiles.org pulling from the same source or is that updated separately?

lukasmartinelli:
> Yeah apologies I've been a bit slow on that I'll try to get on to that tomorrow. On yet another related note @lukasmartinelli can you make @pjsier a member of the maputnik org

Done.

chriswhong:
Hello all, we are a new dev shop at the New York City Department of City Planning.  We're building a lot of map-heavy applications, and maputnik has been a huge help.  You can see our first project that uses mapboxGL here: [https://communityprofiles.planning.nyc.gov](https://communityprofiles.planning.nyc.gov)

[...]

PR #271, opened 2018-03-16
Adding mac keyboard bindings

ziveo:
This PR is adding missing mac keyboard bindings


orangemug:
Hi @ziveo thanks for the pull request, much appreciated. How about using something like

```
const isMac = navigator.platform.toUpperCase().indexOf('MAC')>=0;
if(isMac) {
  // Mac bindings...
}
else {
  // Non mac bindings...
}
```

So we only bind for the environment we're interested in?

Ref: https://stackoverflow.com/questions/10527983/best-way-to-detect-mac-os-x-or-windows-computers-with-javascript-or-jquery

orangemug:
Ah actually it looks like we can just use `mod`. See https://github.com/ccampbell/mousetrap/blob/fec446031c88cfcbd4cbdbfcc6c88f16fae4a50b/mousetrap.js#L135

ziveo:
I think this will work

orangemug:
Thanks @ziveo merging

orangemug:
...also @ziveo if you have time please share what you're using Maputnik for over in ticket #164

PR #278, opened 2018-04-09
Updated deps

orangemug:
Updating a bunch of deps


orangemug:
This isn't an actual failure. It's just because we're using phantomjs for our tests, that browser is old and  deprecated now.

I've got a full new set of webdriver tests almost ready over in <https://github.com/maputnik/editor/compare/master...orangemug:fix/web-driver-tests-v7>

oterral:
@orangemug did you plan to use greenkeeper.io to update dependencies automatically? it's really a nice tool.

orangemug:
Hey @oterral yeah that was one of my motivations behind for having decent tests (see #280). I think we're probably a little way off decent tests still though... However on pull request you also now get an app build (as a build artifact), so that means a PR raised from a service will still allow for manual testing via that build artifact, no need for manual checkout/install.

I'm going to manually updated the dependencies today, then add a service to the repo.

With regards to tooling I was thinking <https://renovateapp.com> after listening to <https://changelog.com/podcast/289>. Is there any chance you've used it?

orangemug:
@oterral also if you have any experience from using greenkeeper.io. What are the things it does well and not so well. Any gotchas or annoyances that you've experienced. Any advice would be much appreciated, as I've not used either.

oterral:
@orangemug  Yes I've seen the PR with tests that's  nice. Tests are so important.

Currently we use greenkeeper.io since 3 months and I have only good things to say so far, good PR descriptions, stable, easy to use .  But it's only for npm dependencies.  I didn't know about renovate but seems quite powerful especially to manage more than npm module dependencies.

PR #281, opened 2018-04-13
Block the popup on click in inspect mode

oterral:
As discussed [here](https://github.com/lukasmartinelli/mapbox-gl-inspect/pull/12)

With that we have the same behavior as Mapbox studio.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7597,9 +7597,9 @@
       }
     },
     "mapbox-gl-inspect": {
-      "version": "1.3.0",
-      "resolved": "https://registry.npmjs.org/mapbox-gl-inspect/-/mapbox-gl-inspect-1.3.0.tgz",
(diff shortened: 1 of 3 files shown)


orangemug:
Hi @oterral thanks for the pull request and welcome to Maputnik.

Clarification for what I was failing at articulating in the other ticket. The issue I have with inspect mode at the moment is that there is just too much data to fit in a tooltip. So the tooltip ends up overflowing off the bottom of the screen (see screenshot). I had a chrome web inspector 'select element' UX in mind.

screen shot 2018-04-13 at 14 51 47

However all that said your solution is substantially better than the current functionality :)

So thanks, merging PR.

oterral:
thx @orangemug . and thx for your efforts on this project It's really amazing.


I'm not sure to understand what do you mean by a  "chrome web inspector 'select element' UX " , cna yu explain?

orangemug:
> ...and thx for your efforts on this project It's really amazing.

Thanks thats really appreciated

Basically something similar to how the select element button in chrome devtools operates. Highlighted in red in the screenshot below.

screen shot 2018-04-13 at 15 51 16

The tooltip would then become a static panel somewhere in the UI.

Thoughts?

PR #284, opened 2018-04-15
Handle data functions without zoom

### LINKED ISSUE #276: Interpolate values
It would be nice to be able to add/edit exponential values like these:
```
"paint": {
    "line-color": {
      "property": "basin",
      "stops": [
        [-300, "#66d4ff"],
        [-200, "#222222"],
        [-100, "#6687ff"],
        [ -99, "#bf66ff"]
      ],
      "type": "exponential"
    },
```

If such code is added manually in json editor, maputnik displays values (colours in this example):
paveikslas

But I'm unable to see/change stop value and even trying to change colour gives an error:

> layers[2].paint.line-color.stops[0]: object stop key must have zoom
> layers[2].paint.line-color.stops[1]: object expected, number found

This is very usefull when creating different thematic maps. For example river basin maps where one property contains basin number and you want to visualise different basins with different colours without creating a different layer for different basin (which could mean a lot of layers).

pjsier:
Fixes #276. I haven't extensively tested this yet, but I wanted to see if this approach makes sense in general, and I can take a closer look if it does.

Modifies the data property field to only creates the zoom input if the provided value is an object, otherwise ignores it and allows the data value line to take the full width. Also modifies `changeStop` to return just the `value` property if `zoom` is undefined.

This should handle any styles imported with the array syntax and without zooms, but still defaults to creating stops with zoom values attached when the data function button is selected.

### DIFF
--- a/src/components/fields/_DataProperty.jsx
+++ b/src/components/fields/_DataProperty.jsx
@@ -51,7 +51,8 @@ export default class DataProperty extends React.Component {
 
   changeStop(changeIdx, stopData, value) {
     const stops = this.props.value.stops.slice(0)
-    stops[changeIdx] = [stopData, value]
+    const changedStop = stopData.zoom === undefined ? stopData.value : stopData
+    stops[changeIdx] = [changedStop, value]
     const changedValue = {
       ...this.props.value,
       stops: stops,
@@ -75,8 +76,8 @@ export default class DataProperty extends React.Component {
     }
 
     const dataFields = this.props.value.stops.map((stop, idx) => {
-      const zoomLevel = stop[0].zoom
-      const dataLevel = stop[0].value
+      const zoomLevel = typeof stop[0] === 'object' ? stop[0].zoom : undefined;
+      const dataLevel = typeof stop[0] === 'object' ? stop[0].value : stop[0];
       const value = stop[1]
       const deleteStopBtn = <DeleteStopButton onClick={this.props.onDeleteStop.bind(this, idx)} />
 
@@ -94,8 +95,9 @@ export default class DataProperty extends React.Component {
         dataInput = <NumberInput {...dataProps} />
       }
 
-      return <InputBlock key={idx} action={deleteStopBtn} label="">
-        <div className="maputnik-data-spec-property-stop-edit">
+      let zoomInput = null;
+      if(zoomLevel !== undefined) {
+        zoomInput = <div className="maputnik-data-spec-property-stop-edit">
           <NumberInput
             value={zoomLevel}
             onChange={newZoom => this.changeStop(idx, {zoom: newZoom, value: dataLevel}, value)}
@@ -103,6 +105,10 @@ export default class DataProperty extends React.Component {
             max={22}
           />
         </div>
+      }
+
+      return <InputBlock key={idx} action={deleteStopBtn} label="">
+        {zoomInput}
         <div className="maputnik-data-spec-property-stop-data">
           {dataInput}
         </div>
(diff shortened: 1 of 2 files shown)


orangemug:
Hmmm, you should also be getting Circle CI builds. @pjsier is there any chance you could push a new commit to the branch, I've just changed the Circle CI settings to allow this (well hopefully anyway)

pjsier:
Just rebased and pushed again and it looks like the CircleCI tests are running

orangemug:
@pjsier thanks. Fun fact, now you get pull request builds, see the artifacts tab here <https://circleci.com/gh/maputnik/editor/40#artifacts/containers/0>

Here is the build for your PR <https://40-67726921-gh.circle-artifacts.com/0/tmp/circle-artifacts.0Fb5Rru/build/index.html>

orangemug:
Test lint failure

```
src/styles/_zoomproperty.scss
 134:3  ✖  Expected empty line before nested      rule-nested-empty-line-before
           rule
```

orangemug:
@pjsier a quick question. So as you said you're only removing the `zoom` field if the JSON doesn't have that data, so there is no way to create that state from the UI. I guess with the the UX in the pull request as it currently stands, it wouldn't be obvious to the user how they would create data functions to be zoom independent. Well without the hack anyway <https://github.com/maputnik/editor/issues/276#issuecomment-379725503>

Is there a UI that would make sense to allow the creation of 'data functions without zoom'. Thoughts?

pjsier:
@orangemug yeah, that's the main part I'm not sure of here. We might be able to add an additional select input to the data function fields, but there's already a lot going on there. On top of that, I'm not sure how much it's worth adding more support for the function syntax when the style spec is really moving to expressions.

I'm leaning toward not adding a way to set this up via the UI and instead just make sure imported styles without zooms are editable, but I think adding a control for it in the UI would be doable

orangemug:
@pjsier I wonder if we should just default zoom to `0` then? The situation I'm imagining is they have a base style which uses `[-300, "#66d4ff"]` syntax then the user can't add zoom levels (`[{"zoom": 0, "value": -300}, "#66d4ff"]`) without editing the JSON?

I guess then though opening that style in Maputnik would alter the style on load.

pjsier:
@orangemug If the plan is to only support the syntax if it's present in an imported base style, it seems like we shouldn't modify the style on load. They would have to modify the JSON or clear out the field and re-add it to add zoom levels, but I'm not sure how common a use case that would be

orangemug:
@pjsier so I think it'd be nice to be able to create value without zoom in the UI. That's what was initially intended over in #276. However I think lets merge this anyway, it'll mean we now support the spec fully again.

It's still marked _WIP_ is there anything that still needs doing or shall I merge?

pjsier:
Works for me! Maybe we could create a separate issue for that? And I want to go through this again and make sure I don't run into any bugs, but once I do that I'll update this PR

orangemug:
> Maybe we could create a separate issue for that?

Makes sense to me

> And I want to go through this again and make sure I don't run into any bugs, but once I do that I'll update this PR

I'll merge when you're ready. I'll cut a new release when this is merged with all the new features/fixes

pjsier:
@orangemug this should be good to merge!

orangemug:
Thanks @pjsier merged!

By the way congrats on https://evictionlab.org/map it looks great.

pjsier:
Thanks! Maputnik was a huge help in styling the basemap and getting the choropleth and bubble layers set up

orangemug:
> Maputnik was a huge help in styling the basemap and getting the choropleth and bubble layers set up

That's great to hear. I would love to hear any feedback, the good/bad and ways we can improve etc...

PR #299, opened 2018-05-10
Added button to expand/collapse layer group

orangemug:
Improves accessibility my making layer expand/collapse keyboard accessible.

**Needs further research:** Should the whole element be a button or just the expand/collapse icon?


orangemug:
My current thinking is this would better be solved with using https://github.com/springload/react-accessible-accordion

orangemug:
Closing issue, see https://github.com/maputnik/editor/pull/299#issuecomment-389813467

PR #306, opened 2018-05-22
Keyboard accessible layer options

orangemug:
Added additional menu into `<LayerEditor/>`. This is to make the following options accessible to keyboard users

 - reorder layers (previously only via drag)
 - duplicate layer
 - delete layer
 - hide/show layer

I've hidden the icons (delete/duplicate/hide) in the layers list view from the keyboard. The reason being is that I'm aiming to switch to an accessible tree view for the layers list. I believe trying to add in keyboard support for sub items of a treeview would be confusing for the user and difficult to develop. These are now also in the 'vertical ellipsis menu' on the layer pane. 

screen shot 2018-05-28 at 12 32 52

I've also added a header describing the pane (`Layer: [id]`) which I'm unsure about.

Looking for feedback (CC @gregorywolanski, @pathmapper)

Note: The main objective here is accessibility for keyboard users. 

You can see a demo at <https://456-84182601-gh.circle-artifacts.com/0/artifacts/build/index.html>


orangemug:
Merging this PR, any objections to the UI please raise a new ticket.

Issue #328, opened 2018-06-13
Survey


gregorywolanski:
I believe that Maputnik needs more contact with its users.

# The goal

To bring us, the makers of Maputnik, together with the users.

# The prototype

[Prototype of the survey](https://gregorywolanski.typeform.com/to/cPgaSY)

You can fill it out and submit it, it’s just a prototype. I’ll remove the test responses.

# How I want to promote the survey 

1. https://maputnik.github.io/editor
2. https://twitter.com/maputnik
3. https://twitter.com/lukmartinelli (?) CC @lukasmartinelli 
4. The next version of Maputnik?

# How I want to promote the survey inside Maputnik

![1](https://user-images.githubusercontent.com/787518/41369200-884c9ff2-6f44-11e8-84dd-75a07f831f54.png)

![2](https://user-images.githubusercontent.com/787518/41369207-8cfa1548-6f44-11e8-8ca8-5cb7b8e69ba0.png)

![4](https://user-images.githubusercontent.com/787518/41367560-aba4bd04-6f3f-11e8-9fcd-369d68164cc6.png)

![3](https://user-images.githubusercontent.com/787518/41367556-a8668fa0-6f3f-11e8-9ae2-9bee3c0fe96d.png)

The text from the modal window:

> You + Maputnik = Maputnik better for you
> 
> We don’t track you, so we don’t know how you use Maputnik. Help us make Maputnik better for you by completing a 7–minute survey carried out by our contributing designer.
> 
> Take the Maputnik Survey
> 
> It takes 7 minutes, tops! Every question is optional.

# What I need from you

It would be nice to see some answers to the following questions (you don’t have to answer all of them):

**a)** Do you think any of the questions are unnecessary?

**b)** Are there some questions I should add?

**c)** Do the words I chose work?

**d)** Do you want to know anything more about Maputnik users?

**e)** Do you think the method I’ve proposed for promoting the survey in Maputnik is both visible and non–invasive?*

\* The modal window for the survey will appear to users only once.

orangemug:
Really nice survey @gregorywolanski

Answers to the questions

> a) Do you think any of the questions are unnecessary?

No

> b) Are there some questions I should add?

I'm unsure if this should be a new question or a modification of 'What do you use Maputnik for?'. What I'd like to find out it whether people use Maputnik for 

 - datavis
 - base maps
 - web or print
 - private in company or public on the internet

[...]

PR #331, opened 2018-06-25
Survey (#328)

gregorywolanski:
Elements promoting the survey inside Maputnik

### DIFF
--- a/src/config/styles.json
+++ b/src/config/styles.json
@@ -26,8 +26,8 @@
   {
     "id": "osm-liberty",
     "title": "OSM Liberty",
-    "url": "https://rawgit.com/lukasmartinelli/osm-liberty/gh-pages/style.json",
-    "thumbnail": "https://cdn.rawgit.com/lukasmartinelli/osm-liberty/gh-pages/thumbnail.png"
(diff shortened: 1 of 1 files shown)


orangemug:
I'm away on holiday for the next 10 days. I might be a little delayed in reviewing it, I'll try to find the time.

orangemug:
Hey @gregorywolanski the pull request looks great!

A few comments

 - I think clicking 'Take the Maputnik Survey' should close the modal
 - Clicking 'Take the Maputnik Survey' should set the 'survey' cookie so the popup doesn't show again on startup
 - `src/components/modals/SurveyModal.jsx` inline styles should be added to a `.scss` files. I've been using [BEM](http://getbem.com/naming/) syntax for [recent css additions](https://github.com/maputnik/editor/blob/master/src/styles/_layer.scss#L184-L204). So classes would be something like `.modal-survey`, `.modal-survey__container`, `.modal-survey__logo`, `.modal-survey__text-highlight`  
 - `.green` css class is a bit abstract, I suggest using `.modal-survey__text-highlight` (see above)

orangemug:
Also we could use localstorage rather than a cookie. This reduces dependencies for Maputnik

gregorywolanski:
@orangemug Excuse me, I messed up something with git.

PR #343, opened 2018-07-15
Prevent bounce scroll on <body/>

orangemug:
The UI is 100% height so prevent bounce scroll on OSX


orangemug:
Yeah I should probably disable node v8 tests for the time being as they are flakey. Merging anyway

PR #345, opened 2018-07-15
Option to download styles with your own tokens

### LINKED ISSUE #289: Not displaying one's own Tilehosting and Mapbox API key in exporting style.json
I entered both Tilehosting key and Mapbox key and used *Export* function to download export style.json file. The file didn't show my API keys in the style.json file. I can get around by adding it myself but would be nice to have that working.

```
"sources": {
    "openmaptiles": {
      "type": "vector",
      "url": "https://free.tilehosting.com/data/v3.json?key={key}"
```

orangemug:
Fixes #289

### DIFF
--- a/src/components/modals/ExportModal.jsx
+++ b/src/components/modals/ExportModal.jsx
@@ -235,9 +235,23 @@ class ExportModal extends React.Component {
   }
 
   downloadStyle() {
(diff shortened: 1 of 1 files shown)


orangemug:
It would be better if it only displayed inputs for the sources that are in the style. I am building another library to generally manage sources, so I'll add in at a later date.

Issue #155, opened 2017-08-08
Load/save from Github Repo


stevage:
This was tracked in #3, which was partially implemented with Gist support, but no repo saving/loading yet.

orangemug:
I'm going to be working on this over the next week.

orangemug:
Ok so I'm still working on this, there were a few other things that needed doing first. So a quick update as to where this is at.

The working branch is here https://github.com/orangemug/editor/compare/fix/update-deps...orangemug:feature/github-v2

It's basically hacked together in order to prove out the concept, it seems to work great. Here's what works (try to ignore the global variables 😱)

 1. Login with github

![1_open](https://user-images.githubusercontent.com/235915/32567553-7ae04b2a-c4b3-11e7-91e9-037743e22b0e.png)

 2. List all a users repositories
 3. Load a style from a repository (currently tries to load `style.json` in the base of a repository)

![3_show_repos](https://user-images.githubusercontent.com/235915/32567572-86b24dd6-c4b3-11e7-841d-6073be167d47.png)

 4. Commit back to that repository

![4_commit](https://user-images.githubusercontent.com/235915/32567578-8e270f70-c4b3-11e7-8ef9-b4c81af14508.png)

These commits were made through the Maputnik UI using that process <https://github.com/orangemug-test/style-test/commits/master>

On my local machine this is all running inside docker containers via <https://github.com/orangemug/maputnik-services> including the GitHub authentication via [micro-github](https://github.com/mxstbr/micro-github) (see <https://github.com/orangemug/maputnik-services/blob/master/docker-compose.yml#L27-L43>).

I've also started on a _flags_ section it's basically `chrome://flags` for Maputnik. This is so I can release it early to get feedback, although releasing it is a few days off yet (probably towards middle of next week)

Next up, find a place to host [micro-github](https://github.com/mxstbr/micro-github), I'm currently thinking https://zeit.co/pricing but if anyone has any better ideas feel free to share.

As always any feedback is welcome.

stevage:
Sounds great. Couple of quick comments:

- Some users have a lot of repos (I have >100). Listing them all could be problematic.
- Sounds like you're not supporting the workflow of either creating a new repo, or saving to a repo that you didn't load from. Or sending a PR. You might want to spell that out to the user with a message at the appropriate time.

orangemug:
[...]

PR #364, opened 2018-08-22
Reduce bundle size

orangemug:
This is a **work in progress**, mostly complete just requires some testing

 - Use the browsers fetch rather than the request module
 - base64-loader -> raw-loader
 - Remove ol3 for now because it's been broken for a while
 - Removed old GitHub gist support as it's no longer functional
 - Removed Mousetrap as we were only using a small part of the functionality
 - Moved to single js file to make things simplier

Here are some stats:
```
               vendor (before)   app (before)   vendor (after)   app (after)
Stat size:     3.26MB            4.5MB          n/a              3.42MB
Parsed size:   1.95MB            1.9MB          n/a              2.09MB
Gzipped size:  549.64KB          449.93KB       n/a              507.41KB
```
```
Gzipped total size before: 549.64KB + 449.93KB  = 999.57KB
Gzipped total size after: 507.41KB (saving ~49% saving)
```

```
Parsed total size before: 1.95MB + 1.9MB  = 3.85MB
Parsed total size after: 2.09MB (saving ~45% saving)
```

This should also improve startup time also. Local js parse times

Before

```
vendor.js 267.46ms
app.js    242.23ms
total     509.69ms
```

After

```
vendor.js  n/a
app.js     341.22ms (~33% speed up)
```

Full results

 - Before: https://343-67726921-gh.circle-artifacts.com/0/artifacts/build/bundle-stats.html
 - After: https://745-84182601-gh.circle-artifacts.com/0/artifacts/build/bundle-stats.html

Note: that the removed OpenLayers dependency accounted for a major chunk of those savings when we fix and re-enable OpenLayers we should make sure it's loaded externally from main bundle and only loaded when required.


src/components/modals/ExportModal.jsx
```diff
@@ -11,206 +11,8 @@ import Modal from './Modal'
 import MdFileDownload from 'react-icons/lib/md/file-download'
 import TiClipboard from 'react-icons/lib/ti/clipboard'
 import style from '../../libs/style'
-import GitHub from 'github-api'
-import { CopyToClipboard } from 'react-copy-to-clipboard'
 
 
-class Gist extends React.Component {
```
loicgasser: Hi, this code seemed depreciated but the ability to save a style is a nice feature. Is there any plan to add such a feature in the future? For instance, let's say I wanted to save a style to Github for instance.

orangemug: Hi @loicgasser, so this used to exist but then GitHub changed their API removing anonymous gists. Keep an eye on issue #155 for any updates

loicgasser: Thanks for the pointer

PR #366, opened 2018-08-23
Take advantage of transform-class-properties and use arrow functions

loicgasser:
This PR introduces no functional changes but make better use of `babel-plugin-transform-class-properties` plugin that lets you define public class fields. ([Already defined in .babelrc](https://github.com/maputnik/editor/blob/master/.babelrc#L3))

More importantly it lets you define arrow functions in your classes, which by definition will automatically inherit the context of the class.

So I systematically introduced 2 simplifications:

1. Remove the constructor and declare `state` as the class level when possible
2. Use arrow functions to avoid the ugly `.bind(this)` in `render`

I left out `src/components/modals/ExportModal.jsx` since you removed most of the code.

I also cleaned up some unused imports.

I hope you agree with the changes.


orangemug:
Hi @loicgasser thanks for the pull request.

So I've moved towards using [react-autobind](https://www.npmjs.com/package/react-autobind) see <https://github.com/maputnik/editor/pull/364/files#diff-f880b57906a072cdffa62da2c954ddd4R56> in a recent PR. I prefer this PR using `babel-plugin-transform-class-properties`, but I thought you might be interested anyway.

I'll release a `v1.5.0` over the next couple of days and then merge this PR in (along with mine) after that.

loicgasser:
sure I'll remove it and leave only the imports cleanup

loicgasser:
You want to use this commit `https://github.com/maputnik/editor/pull/366/commits/e4de10155365a1e89245c9625c700e9cb02c3465` or only use autobind?

orangemug:
@loicgasser nothing to do from your side. I'm happy with https://github.com/maputnik/editor/commit/e4de10155365a1e89245c9625c700e9cb02c3465

Thanks for the PR

PR #367, opened 2018-08-23
Fix glyph access key for openmaptiles

loicgasser:
At the moment when starting the app we don't see a default map which should be the item at the position 0 in `src/config/styles.json`.
https://rawgit.com/openmaptiles/klokantech-basic-gl-style/master/style.json
is the default style.

This PR fixes the issue.

Is this regexp `mapStyle.glyphs.match(/\.tileserver\.org/)` an outdated one?
Does it make sense to simplify the code this way or are there use cases I am not aware of?

I see in the code that sometimes you're using `;` and sometimes not, what is preferred?

### DIFF
--- a/src/libs/style.js
+++ b/src/libs/style.js
@@ -54,27 +54,27 @@ function indexOfLayer(layers, layerId) {
   return null
 }
 
-function getAccessToken(key, mapStyle, opts) {
-  if(key === "thunderforest_transport" || key === "thunderforest_outdoors") {
-    key = "thunderforest";
+function getAccessToken(sourceName, mapStyle, opts) {
+  if(sourceName === "thunderforest_transport" || sourceName === "thunderforest_outdoors") {
+    sourceName = "thunderforest"
   }
 
   const metadata = mapStyle.metadata || {}
-  let accessToken = metadata['maputnik:'+key+'_access_token'];
+  let accessToken = metadata[`maputnik:${sourceName}_access_token`]
 
   if(opts.allowFallback && !accessToken) {
-    accessToken = tokens[key];
+    accessToken = tokens[sourceName]
   }
 
   return accessToken;
 }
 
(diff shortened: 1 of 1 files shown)


src/libs/style.js
```diff
@@ -83,33 +83,28 @@ function replaceSourceAccessToken(mapStyle, key, opts={}) {
...
-  let changedStyle = mapStyle;
+  let changedStyle = mapStyle
 
-  Object.keys(mapStyle.sources).forEach((tokenKey) => {
-    changedStyle = replaceSourceAccessToken(changedStyle, tokenKey, opts);
+  Object.keys(mapStyle.sources).forEach((sourceName) => {
+    changedStyle = replaceSourceAccessToken(changedStyle, sourceName, opts);
   })
 
-  if(mapStyle.glyphs && mapStyle.glyphs.match(/\.tileserver\.org/)) {
```
orangemug: I think this just needs updating, see <https://github.com/openmaptiles/klokantech-basic-gl-style/blob/master/style.json#L19>

loicgasser: don't you think it's better to replace single occurrences of `{key}` in the URL so that you don't have to maintain hard coded regexp?

orangemug: Ah I see you moved the code into `replaceSourceAccessToken`, my mistake

src/libs/style.js
```diff
@@ -83,33 +83,28 @@ function replaceSourceAccessToken(mapStyle, key, opts={}) {
...
   const changedSources = {
     ...mapStyle.sources,
-    [key]: {
+    [sourceName]: {
       ...source,
       url: source.url.replace('{key}', accessToken)
     }
   }
-  const changedStyle = {
+  let changedStyle = {
```
orangemug: This can stay `const`

loicgasser: yes good catch

PR #377, opened 2018-09-03
Updated regex to tilehosting.com

orangemug:
Partial revert of #367 because it was replacing `{key}` glyphs regardless of what source they were from


orangemug:
AppVeyor looks like it's failing to start (no logs)

loicgasser:
@orangemug have you tried to Open General tab of project settings and increment build number?

orangemug:
I'm not sure I understand what you mean @loicgasser

PR #386, opened 2018-09-20
Added guard to map object

### LINKED ISSUE #382: Uncaught TypeError: Cannot set property 'showTileBoundaries' of null
Chrome on Linux: Version 69.0.3497.92 (64-Bit)

When opening a link to Maputnik with a style paramter **in a new tab**, I get this error quite often. Map and Editor are not displayed. After a refresh (F5) it works.

Example link:
https://maputnik.github.io/editor/?style=https://rawgit.com/maputnik/osm-liberty/gh-pages/style.json
 
error
Edit:
There is also an error meassage regarding the glyphs.

orangemug:
I believe this fixes #382

### DIFF
--- a/src/components/map/MapboxGlMap.jsx
+++ b/src/components/map/MapboxGlMap.jsx
@@ -109,8 +109,10 @@ export default class MapboxGlMap extends React.Component {
       this.state.inspect.render()
     }
 
-    map.showTileBoundaries = this.props.options.showTileBoundaries;
-    map.showCollisionBoxes = this.props.options.showCollisionBoxes;
+    if (map) {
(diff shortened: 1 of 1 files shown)


pathmapper:
@orangemug thanks, I can confirm that this fixes the `TypeError: Cannot set property 'showTileBoundaries' of null` reported in #382.

The map is still not loaded but it looks like this is a different issue (the fallback key for Maptiler is not used for requesting the tiles), so I think this PR is good to merge.

orangemug:
Thanks @pathmapper, I will merge when tests are passing.

PR #392, opened 2018-09-23
Color accessibility UI

### LINKED ISSUE #323: Build UI for color accessibility filters
Build a UI for the color accessibility filters added in issue #312.

Initial thoughts: Change the map/inspect button into a 'menu button' of possible options

 - deuteranopia
 - protanopia
 - tritanopia
 - achromatopsia

I'm picking these because this is what [Color Oracle uses](http://colororacle.org/usage.html) and I don't think we want to overwhelm users with options.

Tasks

 - Suggest solution (if different from above), wait for feedback from users
 - Build either mockup or working UI

Note that we use [react-aria-menubutton](https://github.com/davidtheclark/react-aria-menubutton) else where, see

https://github.com/maputnik/editor/blob/edd09ef585c809e033ecd3b71d2745738cd4583a/src/components/layers/LayerEditor.jsx#L254-L274

orangemug:
Raising this pull request for feedback. This is a first attempt for a color accessibility UI.

My thoughts are that because the map can only be in one state at a time to bundle it in with the Maps Inspect/Map toggle.

screen shot 2018-09-23 at 14 43 02
 
Demo: <https://892-84182601-gh.circle-artifacts.com/0/artifacts/build/index.html#0.36/0/0>
Fixes #323

Thoughts and feedback appreciated

Note: I'll fix failing tests when we're happy with the UI


pathmapper:
Makes sense to me to put it there and I like it :+1: 

Maybe adding tooltips with a short explanation of the different options, like:

- deuteranopia -> red-green color blindness
- protanopia -> red-green color blindness
- tritanopia -> blue-yellow color blindness
- achromatopsia -> total color blindness

pathmapper:
This PR is also a welcome improvement regarding information which map state is active.
Currently you know you are in map mode because you are able to switch to inspect mode and vice versa.

pathmapper:
@orangemug using Chrome on Linux the text color of the options is not distinguishable from the background:

image

PR #398, opened 2018-09-24
Show error if style fails to load and disabled button if input is empty

### LINKED ISSUE #396: Notification if no style was loaded
If no or an invalid URL was inserted before click on _Open URL_, the _Loading style_ modal shows up and immediately closes on its own. There should be a notification to the user that no style was loaded.

Ref: https://github.com/maputnik/editor/issues/393#issuecomment-423851803

orangemug:
Fixes #396

Take a look @pathmapper

### DIFF
--- a/src/components/Button.jsx
+++ b/src/components/Button.jsx
@@ -15,6 +15,7 @@ class Button extends React.Component {
   render() {
     return <button
       onClick={this.props.onClick}
+      disabled={this.props.disabled}
       aria-label={this.props["aria-label"]}
       className={classnames("maputnik-button", this.props.className)}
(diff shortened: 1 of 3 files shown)


pathmapper:
@orangemug looks good in general, only one thing:
If the modal is opened again after a fail the button is not disabled anymore if input is empty.

orangemug:
@pathmapper good spot. The latest commit fixes that issue <https://972-84182601-gh.circle-artifacts.com/0/artifacts/build/index.html>

pathmapper:
@orangemug perfect!

orangemug:
Thanks for testing

Issue #316, opened 2018-06-09
Update webpack from v3 to v4


orangemug:
The announcement blog post seems to have a good overview of what has changed <https://medium.com/webpack/webpack-4-released-today-6cdb994702d4>

Note that we do a few odd things like <https://github.com/maputnik/editor/blob/7dcd6d5552ae436f39d383917fe5c5df89cf88f5/src/libs/mapbox-rtl.js#L3-L11>

We also use a fair few plugins.

PR #408, opened 2018-10-06
Updated dependencies

orangemug:
I've updated

 - the majority of the dependencies
 - webpack from v3 to v4 (#316)
 - `normalizeSourceURL` is now our own version instead of babel compiled from the mapbox-gl code. This was always causing lots of issues updating mapbox-gl between versions, so we'll just keep it updated manually until it's added to the API (see https://github.com/mapbox/mapbox-gl-js/issues/2416) 

This should be ready to review/merge. Please test and tell me if anything is broken that wasn't previously <https://1079-84182601-gh.circle-artifacts.com/0/artifacts/build/index.html>


pathmapper:
Thanks @orangemug, looks good to me :+1:

orangemug:
Yay! thanks @pathmapper. Merging

PR #418, opened 2018-10-19
Delay modal close until blur events have triggered

### LINKED ISSUE #416: Style Settings, Export Style, Sources: Apply changes
Changes are only applied when closing the modal via the close button.

It is also possible to close the modal with a click beside the modal but then changes are discarded.

image

This is confusing (it was at least to me :smile:).

It would be better to make the behavior consistent, so that a closing click beside the modal also applies the changes.

Another option would be to add a "save" button to apply changes and the close button only closes the modal (without applying the changes).

orangemug:
Test build <https://1095-84182601-gh.circle-artifacts.com/0/artifacts/build/index.html>

Fixes #416

### DIFF
--- a/src/components/modals/Modal.jsx
+++ b/src/components/modals/Modal.jsx
@@ -19,6 +19,17 @@ class Modal extends React.Component {
     underlayClickExits: true
   }
 
+  // See <https://github.com/maputnik/editor/issues/416>
+  onClose = () => {
+    if (document.activeElement) {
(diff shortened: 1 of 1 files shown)


orangemug:
Ping @pathmapper

pathmapper:
Works great @orangemug !
Tried all three options to close the modals and in every case changes were applied.

orangemug:
Thanks @pathmapper, merging.

PR #447, opened 2018-10-29
Fix the size of the logo in the survey modal

orangemug:
So it doesn't cause a visual jump when it loads.

### DIFF
--- a/src/styles/_base.scss
+++ b/src/styles/_base.scss
@@ -3,6 +3,7 @@
   src: url('../fonts/Roboto-Regular.ttf') format('truetype');
   font-weight: normal;
   font-style: normal;
+  font-display: swap;
 }
 
 @font-face {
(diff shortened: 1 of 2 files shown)


orangemug:
Is the visual jump from the font loading in and changing the line height?

pathmapper:
Hard to tell, could be related to the `<h1>` font, here is how it looks:

visual_jump

pathmapper:
Adding the following to [_modal.scss](https://github.com/maputnik/editor/blob/master/src/styles/_modal.scss)

```
.maputnik-modal-survey__header {
  max-height: 20px;
 }
```

and adding this class to the `<h1>` tag: https://github.com/maputnik/editor/blob/master/src/components/modals/SurveyModal.jsx#L30 
fixes the visual jump for me.

orangemug:
So the typeface will swap in as it's loaded, browsers handle this different but it's basically a trade off between waiting for the typeface to load vs using the app. I've gone with  [`font-display: swap;`](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/font-display) to prioritise a useable app over a pretty app on first load. The typeface loads last which means for those on slow connections it'll be usable quicker.

PR #448, opened 2018-10-29
Shortcut fixes

orangemug:
- Allow shortcuts to trigger from the shortcut modal
 - Fix inspect/map toggle shortcut

### DIFF
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -74,50 +74,39 @@ export default class App extends React.Component {
     })
 
     
-    const keyCodes = {
-      "esc": 27,
-      "?": 191,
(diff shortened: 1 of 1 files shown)


pathmapper:
`?` for the shortcut menu is not working for the keyboard layout I'm using (need to press `Shift` + `ß` to display `?`). 

Also I wonder what should be exactly the effect of `m` (Focus map)?

pathmapper:
> Could you give it a test let me know if it works.

Yes, it works now :tada: 

> with the map focused you can use arrow keys to move the map and also zoom with +/-

Got it, thanks!

PR #449, opened 2018-10-29
Fixes for 'mapboxgl-popup-tip' border color.

orangemug:
The mapbox-gl popup tip was appearing white in the states, `top-left`, `top-right`,  `bottom-left`, `bottom-right`.

### DIFF
--- a/src/mapboxgl.css
+++ b/src/mapboxgl.css
@@ -14,6 +14,22 @@
   border-top-color: rgb(28, 31, 36);
 }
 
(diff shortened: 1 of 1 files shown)


pathmapper:
Nice find, looks good.

BTW: I really like the available per commit builds with CircleCI.

orangemug:
> I really like the available per commit builds with CircleCI.

Yeah it's really nice isn't it. Thanks CircleCI!

It'd be great if there was a little bot that posted the link as a comment to the PR.

Issue #467, opened 2018-11-13
Custom layer property triggers infinite loop 


meetar:
Hi Maputnik crew – I'm having some trouble running the editor locally.

I have a fork of the apollo-bright style hosted [here](https://meetar.github.io/ap-tilezen-mapbox-styles/src/styles/apollo-bright.json), with the fonts and sprites generated using the instructions in their README and hosted alongside the style, and the `{resource_url}` template expressions in the json replaced by absolute paths.

This can be successfully loaded via url by the current hosted editor: https://maputnik.github.io/editor

I also have a fork of the editor [here]( https://meetar.github.io/maputnik-editor/build/build/), with the `build/build` directory (generated by `npm run build`) pushed to `gh-pages`, which can successfully load all of the example files.

However, when I attempt to load the forked style in my forked editor, the editor gets stuck in an infinite loop of requests, apparently due to continuous `componentDidUpdate` calls, and nothing is drawn. The situation seems to be triggered by an external url included in the json file, eg https://github.com/meetar/ap-tilezen-mapbox-styles/commit/3ea1612b0858c1a64d6f1246e19e8c392dc51440 – when I remove this (and any part of the style which references it) the remainder of the style draws as expected.

`git-bisect` identifies the first commit which exhibits this behavior in this situation as https://github.com/maputnik/editor/commit/26ff9f63bbc5df818ed2ca176277d3376ca63d37 "Merge remote-tracking branch 'upstream/master' into fix/remove-componentWillUpdate".

I have the same trouble when serving the editor locally, whether using `npm start` or the node `http-server` package with the `--cors` flag.

Thanks!

meetar:
I just reran `git bisect`, doing a `rm -rf node_modules && npm install && npm run build` each time, which identified 409f81f0d809a84bd9418c2705f63669c0f245f4 "More componentWillUpdate transition fixes." as the first commit exhibiting this behavior, which I think makes a bit more sense.

meetar:
I have a workaround, using a trick from https://github.com/erikras/redux-form/issues/2629#issuecomment-297268290 – I'll submit a PR as a conversation piece, but this seems like something that shouldn't strictly be necessary :/

pathmapper:
Hi @meetar,

some thoughts:

> This can be successfully loaded via url by the current hosted editor

[...]

PR #468, opened 2018-11-13
shouldComponentUpdate hack

meetar:
For discussion – Checks for differences in props and state - addresses #467


pathmapper:
Works for me.

src/components/map/MapboxGlMap.jsx
```diff
@@ -97,6 +97,16 @@ export default class MapboxGlMap extends React.Component {
     }
   }
 
+  shouldComponentUpdate(nextProps, nextState) {
+    let should = false;
+    try {
+      should = JSON.stringify(this.props) !== JSON.stringify(nextProps) || JSON.stringify(this.state) !== JSON.stringify(nextState);
```
sk1p: This can possibly block updates via `setState` - `this.state` contains non-JSON-serializable data (at least `map`), which means `should` is only true if the props change. If the props are the same, the second part of the ORed expression will be executed and throw an error. Not a problem currently, as nothing is relying on updates via `setState`, but may be come a problem later.

I don't fully understand how to fix it though - maybe don't keep `map`/`inspect` in state, and as a property instead?

PR #469, opened 2018-11-14
make docker simpler

morlay:
assets of frontend is built by the last `RUN`,
so we just need to host the built assets.
in node, python is required, so use python simple server to do this.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,4 +12,5 @@ WORKDIR ${HOME}
 RUN npm install -d --dev
 RUN npm run build
 
-CMD npm run start -- --host 0.0.0.0
+WORKDIR ${HOME}/build/build
+CMD python -m SimpleHTTPServer 8888


orangemug:
That makes more sense, thanks.

Also welcome to Maputnik @morlay and thanks for contributing.

orangemug:
Since we don't use the dockerfile in appveyor I'm assuming appveyor was having a bad day. Merging anyway, hopefully it'll pass... else I'll look into the issue once merged.

pathmapper:
There is something wrong with appveyor in general: https://ci.appveyor.com/project/lukasmartinelli/editor/builds/20372093 (@orangemug could you cancel the test there?)

There was recently an [image update](https://www.appveyor.com/updates/2018/11/09/), hopefully https://github.com/maputnik/editor/pull/470/commits/7004259867741dc793c88b6066449ea1124eba08 helps.

orangemug:
> ...could you cancel the test there?

Unfortunately I don't seem to be able to, there's no UI for me to do that... which is odd. Maybe because @lukasmartinelli is the owner of the CI env still. It's odd though because I can trigger a rebuild

PR #482, opened 2019-01-05
Use StringInput for 'formatted' field type

### LINKED ISSUE #481: Symbol layer: no input element for "Field"
Version: 98ded98

When editing a Symbol layer, no form input is shown for "Field". I tracked this down to [this change in the style spec](https://github.com/mapbox/mapbox-gl-js/commit/b69b928841173fc6412724ee611707fb4ecc52c9), which introduced a new ["formatted" type](https://www.mapbox.com/mapbox-gl-js/style-spec/#types-formatted), which is used for `text-field`.

The doc for `text-field` now reads: ```If a plain `string` is provided, it will be treated as a `formatted` with default/inherited formatting options.```Maybe as a first measure, a `<StringInput />` could be used for `formatted` typed fields to provide the same functionality as before.

sk1p:
Fixes #481 - this is just a hotfix, a more complete fix would need to add support for dynamically adding/editing format parts, as in [this example style](https://github.com/mapbox/mapbox-gl-js/blob/master/test/integration/render-tests/text-field/formatted-line/style.json)

### DIFF
--- a/src/components/fields/SpecField.jsx
+++ b/src/components/fields/SpecField.jsx
@@ -81,6 +81,7 @@ export default class SpecField extends React.Component {
               options={options}
             />
           }
+        case 'formatted':
         case 'string':
           if(iconProperties.indexOf(this.props.fieldName) >= 0) {
             return <IconInput


pathmapper:
Demo: https://1070-67726921-gh.circle-artifacts.com/0/artifacts/build/index.html

pathmapper:
Thank you @sk1p !

PR #510, opened 2019-04-10
Update mapbox dependencies to current version

lukaswelte:
The current editor uses a beta version of mapbox gl js.
We should switch to a stable release so that the [documentation of Mapbox Styles](https://docs.mapbox.com/mapbox-gl-js/style-spec/#paint-line-line-pattern) fits the editor behavior.

For example in line-pattern in this beta data-driven styling is not allowed fully.

Everything else should continue to work as expected

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1253,6 +1253,24 @@
         "wgs84": "0.0.0"
       }
     },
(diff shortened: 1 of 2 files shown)


pathmapper:
Thanks @lukaswelte

Issue #473, opened 2018-11-17
v1.6.0


orangemug:
I've created a project for [v1.6.0](https://github.com/maputnik/editor/projects/2) for the remaining tickets that need doing, pre go live.

If anyone experiences any breaking changes that weren't in the v1.5.0 release, please raise a ticket.

pathmapper:
Things to do:

- [ ] Merge https://github.com/maputnik/desktop/pull/5 after editor release and build CLI binaries.

pathmapper:
Hey @orangemug,

we were asked by @klokan to start using the new Maptiler key (#542) as soon as possible. 

Do you have time to make a new release within a week and deploy it to https://maputnik.github.io/editor/? Otherwise we would run again into https://github.com/maputnik/editor/issues/541 (see https://github.com/maputnik/editor/issues/541#issuecomment-516066660).
 
We've done almost everything on our list https://github.com/maputnik/editor/projects/2 
(and a lot more :rocket:) and I think we could leave the remaining task for later.

I could build the Desktop CLI binaries after the editor release.

orangemug:
@pathmapper sure, I'll build a beta release tomorrow for testing

orangemug:
Release available for testing at <https://maputnik.github.io/releases/v1.6.0-beta/>

Pinging @pathmapper

pathmapper:
Thanks @orangemug, looks good.

orangemug:
Released https://maputnik.github.io/blog/2019/10/08/release-v1.6.0

pathmapper:
Thanks @orangemug, would you mind updating also the release page: 
https://github.com/maputnik/editor/releases ?

orangemug:
> Thanks @orangemug, would you mind updating also the release page:
https://github.com/maputnik/editor/releases ?

Done <https://github.com/maputnik/editor/releases>

pathmapper:
Added desktop CLI binaries to the release: https://github.com/maputnik/editor/releases/tag/v1.6.0

PR #512, opened 2019-04-16
Add support for connecting to a local server on port other than 8000

JesseCrocker:
This works in conjunction with https://github.com/maputnik/desktop/pull/5 to add support for running the desktop app on a port other than 8000, because many people already have something else running on 8000.

### DIFF
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -89,8 +89,15 @@ export default class App extends React.Component {
     autoBind(this);
 
     this.revisionStore = new RevisionStore()
+    const params = new URLSearchParams(window.location.search.substring(1))
+    let port = params.get("localport")
+    if (port == null && (window.location.port != 80 && window.location.port != 443)) {
+      port = window.location.port
+    }
     this.styleStore = new ApiStyleStore({
-      onLocalStyleChange: mapStyle => this.onStyleChanged(mapStyle, false)
+      onLocalStyleChange: mapStyle => this.onStyleChanged(mapStyle, false),
+      port: port,
+      host: params.get("localhost")
     })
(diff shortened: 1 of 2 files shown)


JesseCrocker:
Is there no interest in reviewing this or https://github.com/maputnik/desktop/pull/5 ? If so i'll close them.

pathmapper:
@JesseCrocker thanks for your PRs and sorry for the late response.

I think this is a useful feature that we should implement. Did some testing and it works great so far.

If anyone else wants to try this out, here are the CLI binaries 
(including this PR and maputnik/desktop#5):

- [linux.zip](https://github.com/maputnik/editor/files/3434377/linux.zip)
- [darwin.zip](https://github.com/maputnik/editor/files/3434379/darwin.zip)
- [windows.zip](https://github.com/maputnik/editor/files/3434381/windows.zip)

Since we are using the editor release bundle for building Maputnik Desktop (https://github.com/maputnik/desktop/blob/master/Makefile#L26) I think we should merge this PR before the next release (which will be `1.6`, see https://github.com/maputnik/editor/issues/473, and https://github.com/maputnik/editor/projects/2).

After editor `1.6` is released we'll merge https://github.com/maputnik/desktop/pull/5 and build the new CLI binaries.

@orangemug @lukaswelte any thoughts?

lukaswelte:
@pathmapper that sounds good.

Nice addition @JesseCrocker, thank you :)

pathmapper:
Thanks again @JesseCrocker, left a note in https://github.com/maputnik/editor/issues/473 regarding https://github.com/maputnik/desktop/pull/5.

Issue #488, opened 2019-01-29
Not possible to change numeric values


pathmapper:
From https://github.com/maputnik/editor/issues/487 (thanks @zstadler):

> In the Docker version, it is impossible to type negative and non-integer values in numeric fields . For example, entering -0.5 in the "offset" field is blocked as soon as one the "-" key is pressed.
> 
> It is possible to enter such a number using the "home" or arrow keys. For example "5", "Home", ".", "Home", "-"
> 
> The online version allows typing "0.5" and "-.5" but would not allow typing "-0.5"

Docker version: `master`
Online version: `v1.5.0`

pathmapper:
As reported by @petrsloup in https://github.com/maputnik/editor/issues/484#issuecomment-470141227 there are in general issues with numerical inputs (not only for negative or non-integers values).

This bug was likely introduced with https://github.com/maputnik/editor/pull/358.

pathmapper:
There is a proposed solution ready for testing in #517.

cc @zstadler, @petrsloup

pathmapper:
@orangemug fixed it in #517

PR #517, opened 2019-05-18
Fix for string/number inputs when inputting invalid of intermediate values

orangemug:
Somewhere along the way we (probably me) broke the ability to enter intermediate values into string/number inputs. The editor was constantly trying to replace the value you entered with a 'valid' value.

This resulted in not being able to enter values like `0.1` because `0.` getting automatically 'fixed' to be `0`. Which is obviously not what I wanted 😄 

This change adds an 'editing' state to the inputs, it'll keep updating the map with the latest changes from the inputs, but will prevent the `props` from updating the input until you no longer have focus. 

If anybody has some time to give it a test https://1236-84182601-gh.circle-artifacts.com/0/artifacts/build/index.html#0.34/0/0

### DIFF
--- a/src/components/inputs/NumberInput.jsx
+++ b/src/components/inputs/NumberInput.jsx
@@ -13,25 +13,28 @@ class NumberInput extends React.Component {
   constructor(props) {
     super(props)
     this.state = {
-      value: props.value
+      editing: false,
+      value: props.value,
     }
   }
 
   static getDerivedStateFromProps(props, state) {
-    return {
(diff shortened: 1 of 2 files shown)


pathmapper:
Did some tests, also for all cases mentioned in #488 and it seems to work.

orangemug:
Thanks for testing @pathmapper

zstadler:
It seems fine by my tests.
Thanks @orangemug  for the fix!

PR #518, opened 2019-05-18
Toggle color accessibility based on user agent

### LINKED ISSUE #472: Color accessibility filter browser support
Its turns out the color accessibility filters on WebGL canvases only works on Firefox/Chrome, so we'll need to disable them for other browsers. There doesn't appear to be a way to feature detect this so we'll need to detect the browser currently being used and toggle on that instead.

I suggest keeping the same UI (the new `<select/>` box) and just removing color filters for browsers other than Firefox/Chrome.

screen shot 2018-11-17 at 16 03 41

orangemug:
It's only supported in chrome and firefox and we can't feature detect this unfortunately.

Fixes #472

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1670,7 +1670,6 @@
       "resolved": "https://registry.npmjs.org/align-text/-/align-text-0.1.4.tgz",
       "integrity": "sha1-DNkKVhCT810KmSVsIrcGlDP60Rc=",
       "dev": true,
-      "optional": true,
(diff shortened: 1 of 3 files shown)


pathmapper:
Demo: https://1260-84182601-gh.circle-artifacts.com/0/artifacts/build/index.html

pathmapper:
Looks good, tried it with Chrome, Chromium, Firefox and Edge.

Issue #372, opened 2018-09-01
Performance


gregorywolanski:
So far 3 people (well, 4, but the fourth person didn’t write it directly) mentioned performance in their answers to the question “What annoys you about Maputnik?” in the survey (original spelling):

- performance

- it's a bit slow

- Holy crap... when data sources have a mind of their own for when they are going to show up as options for creating new layers. Sometimes it feels like it takes 10 minutes, sometimes it feels random. They always eventually show up though.

orangemug:
Lets break this ticket down into 2 parts

 - Startup performance
 - Runtime performance

## Startup performance
I don't think startup performance effects us too much, although it'd be nice if Maputnik boot time was quicker it currently takes ~4.5seconds on a 4g connection with an empty cache (I'm on a train currently). This isn't great in the grand scheme of things but I think it isn't a major concern right now. Most of our users will be on a decent internet connection because Maputnik only works on a desktop. From cache it's consistently quick to load.

What is important is that it doesn't get worse, we have tools in our CI to monitor this on every commit. You can look at the artifacts of any commit and see how the bundle size has changed. An example report <https://1075-84182601-gh.circle-artifacts.com/0/artifacts/build/bundle-stats.html>. We'll continue to try and reduce this over time. A prime candidate is only loading the renderer we actually currently require 'mapbox-gl' or 'open layers' (I'll raise a ticket for this later)


## Runtime performance
Runtime performance is pretty poor at the moment, apologies to everyone who has felt this pain. Let's explain what's going on here, why it's an issue and how we're going to resolve it.

Every time you make an edit in Maputnik the style gets updated in the core of the editor and then the following happens

 1. Validate the new style
 2. Update (in react) the UI with the new style
 3. Replace the style in the `mapbox-gl` renderer

The slow bit for us here is item (**2**). We update the entire of the react tree, react then diffs the render tree updating what has changed in the DOM.

On my fairly fast laptop changing the `background` layer's `color` property for the [osm-liberty](https://github.com/maputnik/osm-liberty) takes about 84ms. This is in chrome which is a fair bit faster than Firefox at rendering our UI. The magic number for an app is around 16ms (`1000ms/60fps`). So basically we have some way to go.

[...]

PR #521, opened 2019-05-19
Perf improvements with icepick-style 

orangemug:
I've been busy over the last week trying to improve the Maputnik UI performance. Here's where I've got to so far

I've created a new immutable style store for Maputnik which uses [icepick](https://github.com/aearly/icepick) called [icepick-style](https://github.com/mgljs-contrib/icepick-style)

My simple initial performance test is changing a single color in the UI. In particular, the background layer of the osm-liberty. You can read some further details over in https://github.com/maputnik/editor/issues/372#issuecomment-428677612 

Here are the results so far in the development build of our app. 

Before
Screen Shot 2019-05-19 at 18 29 44

After
Screen Shot 2019-05-19 at 18 28 28

`229.48ms` down to `76.04ms` for 'react tree reconciliation' on a single run, but results seem consistent.

The app seems to mostly work as it did previously. There are a couple of bugs around updating sprites and sources, which I'll try to fix the next time I work on Maputnik.

It's still pretty slow but a big step in the right direction.


pathmapper:
@orangemug this looks promising, but I wonder why _before_ it was `229.48ms` because you mentioned in https://github.com/maputnik/editor/issues/372#issuecomment-428677612:

> On my fairly fast laptop changing the background layer's color property for the osm-liberty takes about 84ms. This is in chrome which is a fair bit faster than Firefox at rendering our UI.

What has changed since then? Different machine? browser? or did you measure there something different than 'react tree reconciliation'?

orangemug:
> What has changed since then? Different machine? browser? or did you measure there something different than 'react tree reconciliation'?

So the two measurements are the same test in different versions of the code with the same browser/laptop

 - `229.48ms` was tested in `v1.5.0` of Maputnik without the immutable additions
 - `76.04ms` is with my latest immutable style changes applied

It's a pretty simple test at the moment, but essentially the use of immutable data structure and  `PureComponent` means that react doesn't bother calling the render methods for a bunch of the component tree. Which in turn means that there is a lot less work for the 'react tree reconciliation' to do.

Also worth noting that this is just the first step, we should be able to improve this significantly more. Saying that my first priority is getting these current improvements stable and well tested and merged into `master`

pathmapper:
Thanks, I assumed you were using `v1.5.0` for https://github.com/maputnik/editor/issues/372#issuecomment-428677612 and achieved `84ms` without the immutable additions.

orangemug:
> Thanks, I assumed you were using v1.5.0 for #372 (comment) and achieved `84ms` without the immutable additions.

Nice spot @pathmapper I'll checkout the earlier versions of Maputnik see if the perf has dropped. It's possible I just measured it incorrectly in https://github.com/maputnik/editor/issues/372#issuecomment-428677612 but I'll do some hunting about and see what I can find out.

PR #522, opened 2019-05-20
Add range input for minZoom/maxZoom

### LINKED ISSUE #432: Zoom slider
A user suggested to replace “Min Zoom” and “Max Zoom” with two or one slider.

orangemug:
As described in #432 I've added `<input type="range/>` as new interface for entering `minZoom`/`maxZoom`

Demo: <https://1314-84182601-gh.circle-artifacts.com/0/artifacts/build/index.html>

Fixes #432


pathmapper:
Thanks @orangemug , with Chrome it works nicely, with Firefox it's not possible to slide the handles (they get resetted immediately).

pathmapper:
@orangemug maybe the Firefox issue is related to a race condition:
https://stackoverflow.com/questions/30550314/react-input-range-slider-not-propagating-last-value-to-its-owner-in-chrome (in this post Chrome isn't working)

orangemug:
> @orangemug maybe the Firefox issue is related to a race condition:
https://stackoverflow.com/questions/30550314/react-input-range-slider-not-propagating-last-value-to-its-owner-in-chrome (in this post Chrome isn't working)

@pathmapper I think you're partially correct, however I fixed the race condition and it was still failing. The only thing I can think is that the 'UI pause' (because of the slow performance) when the DOM updates causes the range slider in FF to 'halt'.

I've changed the code to only update the value when the `pointerup` event fires. I think this is the best we can do at the moment, what do you think?

Demo <https://1336-84182601-gh.circle-artifacts.com/0/artifacts/build/index.html#0.34/0/0>

pathmapper:
@orangemug I think it's fine to only update the value on `pointerup`.

However, there is still something wrong (FF 67.0 on Linux):

Most of the times, if the mouse is moved after `pointerup`, the slider gets resetted:

ff_slider

orangemug:
Hmmm, well this is a pain, I don't get that issue on FF on OSX. I'll try to sort out a VM for testing see if I can get to the bottom of it.

pathmapper:
Here is the FF on Linux issue with open _Components_ tab of the latest CI profiling build. Including `PointerUp` everything works as expected. The issue appears if the mouse if moved after `PointerUp`: The value resets and the state switches to `editing: true`.

range_slider_ff

pathmapper:
Nice try, but the issue is still there.

Maybe this helps (we are using `display: flex`):
https://stackoverflow.com/questions/57225124/react-vertical-range-slider-always-resets-to-default-ignoring-new-state

Or use something like this:
https://www.npmjs.com/package/react-rangeslider

pathmapper:
Here a gif with the logging which was added in the last commit:

range_slider_ff_logging

orangemug:
Thanks @pathmapper, that's helpful. I've added an additional attempt at a fix and some more logging.

pathmapper:
We're getting close! 

The previous issue is fixed (the value is not resetted anymore if the mouse is moved after `PointerUp`):
range_slider_2




Now the value is resetted only if there is **no** mousemove after `PointerUp` before dragging the slider again:
range_slider_3

pathmapper:
The issue is still there, the logs for `getDerivedStateFromProps` are `undefined`

range_slider_4

orangemug:
`maxZoom` logs itself as `getDerivedStateFromProps[1]` I've move some logging about so it actually logs that value.

pathmapper:
New logs:

range_slider_5

pathmapper:
This is much better @orangemug, however there is still sometimes a jumping, current logs:

Peek 2020-01-19 10-01

pathmapper:
Noticed another thing. If you open a new style, the sliders are not showing the actual min/max zoom of this new style. They are still showing the values of the style before. This is also an issue with `v1.6.1` but not with `master` anymore. Maybe update this branch to `master`. If we are lucky the jumping gets also away :thinking: 

Peek 2020-01-19 10-02

orangemug:
Works in development not once compiled via CI... bear with me something strange is going on, investigating

orangemug:
https://1956-84182601-gh.circle-artifacts.com/0/artifacts/build/index.html#0.36/0/0 this is looking better, I've tested on

 - chrome, firefox - OSX
 - chrome, firefox - XUbuntu

I managed to recreate your original bugs about 1/20 times locally in a VM, and can't now with the changes. So hopefully this fixes the issues 🤞

pathmapper:
:tada: :tada: :tada: 

No issues any more on my side, works great!

orangemug:
Nice, thanks for all the help @pathmapper, amazing test support!

Merging! 🎉

PR #529, opened 2019-06-16
OpenLayers projection support

orangemug:
This pull request is the first step towards OpenLayers projection support.

_The motivation_: I wanted to create a vector map of Antarctica, and couldn't find an easy way to do that. I found a few bit a pieces here and there but the examples I found used 'dirty reprojection' (see https://medium.com/devseed/dirty-reprojectors-1df66e8f308d). Which I wasn't really a fan of because it messes up the lat/lon of the map.

Because OpenLayers has proper projection support (mapbox-gl does not), I thought I have a go at creating some vector tiles in different projections and adding a UI to Maputnik. Here's what I've got so far.

I have a script which generates vector tiles using GDAL from a definition file https://github.com/orangemug/vector-tiles-projection-test/blob/master/definition.json . The definition files specifies a target projection (`epsg`) with GDAL auto detecting the source projection.

```
    {
      "id": "antarctic_coastline",
      "name": "",
      "attribution": "",
      "description": "",
      "epsg": "3031",
      "min_zoom": 0,
      "max_zoom": 2,
      "out_dir": "vectortiles/EPSG3031",
      "input_file": "data/EPSG3031/Coastline_medium_res_polygon/Coastline_medium_res_polygon.shp"
    }
``` 

I've had to make some small changes to `ol-mapbox-style` to load the projection from the tilejson https://github.com/openlayers/ol-mapbox-style/compare/master...orangemug:feature/add-projections . I'll raise a ticket in the `ol-mapbox-style` repo to try and get these merged over the next couple of weeks. I'm sure there will be changes required because I'm pretty new to OpenLayers.

The styles have 2 new keys in the source `projection` and `extent` https://github.com/orangemug/vector-tiles-projection-test/blob/2df924154f7f88b59fb7543d47a6b9972ffc601b/styles/epsg3031.json#L9-L15 which are loaded by the modified `ol-mapbox-style` described above.

There is also some additional UI in the settings panel

Screen Shot 2019-06-16 at 22 05 52

Here's a screenshot of an antarctic map in Maputnik

Screen Shot 2019-06-16 at 22 06 30

Also a couple of links to the demos

 - [https://orangemug.github.io/vector-tiles-projection-test/styles/epsg102003.json](https://1406-84182601-gh.circle-artifacts.com/0/artifacts/build/index.html?style=https://orangemug.github.io/vector-tiles-projection-test/styles/epsg102003.json)
 - [https://orangemug.github.io/vector-tiles-projection-test/styles/epsg3031.json](https://1406-84182601-gh.circle-artifacts.com/0/artifacts/build/index.html?style=https://orangemug.github.io/vector-tiles-projection-test/styles/epsg3031.json)

So how can you all help

 - Look at the demos and tell me if anything isn't working?
 - Is the general approach for generating vector tiles sensible? Am I missing anything obvious?
 - What do you all think?

Note: This is probably going to be sometime before it's merged, I wanted to get it out there so I can gather feedback and to give me time to try and get the ol-mapbox-style changes upstream. I'm going to switch back over to #521 in the short term to try and improve the performance of Maputnik but this will be moving along slowly in the background.


pathmapper:
@orangemug this looks very interesting!

 Some questions and thoughts (I'm not familiar with ol projections until now):

- Is the additional source key `projection` really needed? The projection is already in the style metadata (could be named `ol:projection`). If there are multiple sources in a style, the projection must be the same for all sources, right? 

- What's the purpose of the source key `extend`? Could `bounds` (https://docs.mapbox.com/mapbox-gl-js/style-spec/#sources-vector-bounds) used instead?

orangemug:
Good questions @pathmapper

OpenLayers can do 'client-side raster reprojection' of tiles https://openlayers.org/en/latest/examples/reprojection-by-code.html. So in that instance you could in theory have a projection on a source that is different to that of the map. This is only available for raster sources currently however

 - I thought it was good to keep the API consistent across sources
 - I could see this being available for vector sources in the future

It's worth noting that as far as I can see client side reprojection has some drawbacks, so it's not a fix all. So either way most of the time you'd want tiles in the maps intended projection anyway.

Also worth noting, that at the moment the changes for `ol-mapbox-style` are only for vector sources. It'd be good to have raster also, but that's for later down the road.  

The `ol:extent` is the extent in the coordinate reference system of the projection. Using lat/lon gets messy with polar projections. Trying to type this reasoning I'm having problems putting it into words, but basically the south pole is `{lat: -90, lon: ??}`, where `lon` can be anything from `-180` - `180`, things just seems to get complex really quickly.

I'm pretty novice on all this also, so if anything sounds incorrect it might well be.

Issue #129, opened 2017-04-11
Text layout properties - add/remove font


pathmapper:
For adding optional fallback font(s) for a layer, one has to use the JSON editor at the moment. Maybe it would be possible to implement it directly in "Text layout properties", e.g. a "+" button that adds 
```
,
""
```
to the JSON style.

A "&#8722;" button could be used to delete an existing fallback font.
Current situation: When deleting the name of a fallback font in "Text layout properties", then 
```
,
""
```
remains in the JSON style.

stevage:
Another motivating problem: every symbol layer has two fonts by default, and you can't remove the second one. That means you can't use orangemug's [excellent Github repo](https://github.com/orangemug/font-glyphs-v2/tree/gh-pages) because it doesn't support fetching multiple fonts at once (".../Open Sans,Arial Unicode/....").

(I'm not really clear why one needs fallback fonts, when all the glyphs are being fetched from the same server - why would the first one fail?)

orangemug:
@stevage it's not really fallback fonts but rather fallback characters (glyphs). Its a little confusing let me explain.

Typefaces only support a part of the unicode character range, this support varies I believe https://www.google.com/get/noto is the only typeface to support it all. So when we are saying `Open Sans,Arial Unicode` what we're looking for is `Open Sans` if the character isn't supported then the glyph for `Arial Unicode` will be returned instead.

However the client (mapbox-gl-js) knows nothing about this but instead just sends a request for the font stack and the glyph range it's looking for. An example request might be `http://example.com/fonts/Open Sans,Arial Unicode/0-255.pbf` the server that deals with that request merges the glyphs from the font stack in the response.

There is support for this in https://github.com/klokantech/tileserver-gl and for more info on how they did that look at https://github.com/klokantech/tileserver-gl/blob/master/src/utils.js#L104

orangemug:
As for the issue, although it would be nice to have multiple fonts supported until we have a public server that implements the above we are unable to implement this feature.

Note: For future reference we might be able to do an ugly hack once the [Service Worker API](https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API) is widely supported by intercepting the request.

pathmapper:
> until we have a public server that implements the above we are unable to implement this feature.

[Maptiler](https://www.maptiler.com/cloud/) supports this: 
- https://maps.tilehosting.com/fonts/{fontstack}/{range}.pbf?key={key}

And there is also a hack available for hosting from static directory:
[...]

Issue #509, opened 2019-04-06
Ability to delete properties completely


pathmapper:
Currently the only way to delete a property completely is to use the JSON editor.

It would be usefull to be able to do it also in the UI somewhere near the input fields.

Example use cases:
- https://github.com/maputnik/editor/issues/129, https://github.com/maputnik/editor/issues/422 (`text-font`)
- https://github.com/maputnik/editor/issues/507 (`fill-pattern`)

pathmapper:
Fixed in #531

Issue #395, opened 2018-09-23
Font change breaks rendering and map interaction needed to update map view


pathmapper:
Style: https://maputnik.github.io/editor/?style=https://rawgit.com/maputnik/osm-liberty/gh-pages/style.json

Glyph URL: `https://orangemug.github.io/font-glyphs/glyphs/{fontstack}/{range}.pbf`
(same behaviour for: `https://fonts.openmaptiles.org/{fontstack}/{range}.pbf`)

![change_font](https://user-images.githubusercontent.com/20856381/45932166-b34dc800-bf78-11e8-948e-e78299f2f2f1.gif)

For comparison:

Style:  https://maputnik.github.io/editor/?style=https://raw.githubusercontent.com/openmaptiles/klokantech-basic-gl-style/master/style.json

Glyph URL: `https://free.tilehosting.com/fonts/{fontstack}/{range}.pbf?key={key}`
Looks like a fallback font is served even if there is no fallback font referenced in the style for this layer.

![change_font_tilehosting](https://user-images.githubusercontent.com/20856381/45932482-1a6d7b80-bf7d-11e8-9eb4-5c5dfec8dedf.gif)

orangemug:
So looking at the web inspector in chrome, it looks like `https://free.tilehosting.com/fonts/` always returns a http 200 status, regardless of it the font name is valid or not. Whereas `https://orangemug.github.io/font-glyphs/glyphs/` 404's

I think we need to

 1.  Update to `mapbox-gl` to 0.49.0 (latest) see if the problem is still occurs
 2. If the issue still occurs, raise a bug in `mapbox-gl`

pathmapper:
After update `mapbox-gl` to 0.49.0 and `mapbox-gl-style-spec` to 13.2.1, I get the following errors 
on `npm install` :

```
ERROR in ./node_modules/mapbox-gl/src/geo/coordinate.js
Module parse failed: Unexpected token (13:10)
You may need an appropriate loader to handle this file type.
|  */
| class Coordinate {
|     column: number;
|     row: number;
|     zoom: number;
 @ ./node_modules/mapbox-gl/src/util/util.js 51:18-46
 @ ./node_modules/mapbox-gl/src/util/mapbox.js
 @ ./src/components/App.jsx
 @ ./src/index.jsx
[...]

PR #531, opened 2019-06-20
Ability to delete properties completely

### LINKED ISSUE #507: Regression: Map doesn't update when fill-pattern is changed

orangemug:
This fix allows you to remove properties which display as `AutocompleteInput`, `ColorField` and `InputBlock`, where as before they would be set to an empty string `""`   

Demo: <https://1416-84182601-gh.circle-artifacts.com/0/artifacts/build/index.html>

Fixes #507
Reference #509 (`text-font` still requires further fixes)


pathmapper:
This doesn't work (map is not updated) if you modify (don't delete) e.g. the name of a  `fill-pattern` in the `pattern` field or in the JSON editor.

Also after a property (tested with `fill-pattern`) is deleted completely and you type a valid value in the `pattern` field, the map is not updated. The map is only updated if a proposed value in the `pattern` field is selected.

It's not possible anymore to set an empty string as value via the `pattern` field. The only option is to use the JSON editor.

orangemug:
> This doesn't work (map is not updated) if you modify (don't delete) e.g. the name of a fill-pattern in the pattern field or in the JSON editor.

Yeah I agree but I think this is an existing issue. Try in the current build of `master` https://1583-67726921-gh.circle-artifacts.com/0/artifacts/build/index.html

> Also after a property (tested with fill-pattern) is deleted completely and you type a valid value in the pattern field, the map is not updated. The map is only updated if a proposed value in the pattern field is selected.

I think this is also an existing issue in `master`

> It's not possible anymore to set an empty string as value via the pattern field. The only option is to use the JSON editor.

So I thought this was what was originally reported in #507, and the intended functionality when would you want an empty string?

pathmapper:
> Yeah I agree but I think this is an existing issue. Try in the current build of master https://1583-67726921-gh.circle-artifacts.com/0/artifacts/build/index.html

Right, this is an existing issue in `master` and not introduced by this PR, but in `1.5` the issue is not there. Here's a `fill-pattern` test case:
https://maputnik.github.io/editor/#17.96/50.775437/6.084083 (`OSM Liberty` style, `road_area_pattern` layer)

> I think this is also an existing issue in `master`

Right, but also in `1.5` not present.

> So I thought this was what was originally reported in #507 (although the title is a little confusing), and the intended functionality when would you want an empty string?

Sorry for the confusion. I understood that in #507 originally the regression since `1.5` was reported (in `1.5`, if you change a `fill-pattern` with a valid value to an empty string the pattern is not shown anymore on the map) and later the ability to delete properties completely was requested.

So my intend was that #507 is about the regression since `1.5` (as `1.5` is working fine) and #509 is about the possibility to delete a property completely (maybe by adding a button for this task next to the input field).

Regarding empty string as value:
1. Empty string is a valid value for `fill-pattern` according to the style spec (it says it's a string, and empty string is a string).
2. `fill-pattern` disables other properties (e.g. `fill-color`), so a use case is to create a style with empty string as `fill-pattern` value and set the `fill-pattern` value later programatically in an app with https://docs.mapbox.com/mapbox-gl-js/api/#map#setpaintproperty

orangemug:
Hey @pathmapper thanks for the feedback.

So regarding the updating issues with `fill-pattern` downgrading `mapbox-gl` to `^0.47.0` fixes the issue, so it look like the problem is upstream. I'll see if I can get a test case outside this PR. 

As for the `fill-pattern` empty string (`""`) I feel like is a bit of an edge case. Do people use it like that, because you could achieve the same thing other ways. For example setting `{"layout": {"visibility": "none"}`, and then enabling the layer when you need it.

Also remember that initially the text box is blank (with a new layer), but the JSON won't have `{"fill-pattern": ""}` until you
 
 1. Enter something in the `<input/>`
 2. Remove the text again
 3. Now the JSON will be set to `{"fill-pattern": ""}` 

Since the `fill-pattern` as empty string seems like a bit of an edge case. I think we should prioritise removing the value if the `<input/>` is `""`. If a user really needs `""` they can always add it back in the the JSON editor.

Thoughts?

pathmapper:
> regarding the updating issues with fill-pattern downgrading mapbox-gl to ^0.47.0 fixes the issue, so it look like the problem is upstream.

Good find @orangemug ! Looks like this is fixed in mapbox-gl `v1.1.0-beta.1` :smile: 

> Also remember that initially the text box is blank (with a new layer), but the JSON won't have {"fill-pattern": ""}

Very good point. 

Let's go for "removing the value if the `<input/>` is `""`. If a user really needs `""` they can always add it back in the the JSON editor."

orangemug:
> Looks like this is fixed in mapbox-gl v1.1.0-beta.1

Nice spot @pathmapper 

Updated to `mapbox-gl@^1.1.0-beta.1` heres the new build <https://1432-84182601-gh.circle-artifacts.com/0/artifacts/build/index.html>

pathmapper:
Did some tests and it works great so far. :tada:

pathmapper:
@orangemug it's not possible to delete a font value in the `Font` field anymore, but I guess this is what you've meant with 

> Reference #509 (text-font still requires further fixes)

grafik

orangemug:
> it's not possible to delete a font value in the Font field anymore, but I guess this is what you've meant with

Thanks @pathmapper it was what I meant... but I didn't quite realise how badly I'd broken it. I've push up a new improved `<FontInput/>` it basically removes empty values and always has a empty input at the bottom of the list for entering additional fallback fonts.

It's still a little awkward to use but seems better than before (`<FontInput/>` is probably up for a rewrite soonish). Let me know what you think. 

Link: <https://1445-84182601-gh.circle-artifacts.com/0/artifacts/build/index.html>

pathmapper:
Thanks @orangemug , looks good!

I think like it is now this PR closes also https://github.com/maputnik/editor/issues/129 and https://github.com/maputnik/editor/issues/509.

orangemug:
Yay merging! Thanks for all the help @pathmapper really great work testing this. 

Fixes #129 and #509.

pathmapper:
Fixed also #395 :tada:

PR #546, opened 2019-08-03
Optimize docker image

nyurik:
* Use 2 stage docker building to produce a tiny python3-slim based docker image with just the compilation results.


README.md
```diff
@@ -26,6 +26,13 @@ Mapbox has built one of the best and most amazing OSS ecosystems. A key componen
 ## Donations
 If you or your organisation has seen value from Maputnik, please consider donating at <https://maputnik.github.io/donate>
 
+## Using
```
pathmapper: There are already two ways mentioned in the readme how the editor could be used (online at https://maputnik.github.io/editor/ and CLI). I think it's good to have this new section _Using_ and we should move the other two options in here. I would like to keep the order: 
1. online editor
2. CLI
3. docker

README.md
```diff
@@ -26,6 +26,13 @@ Mapbox has built one of the best and most amazing OSS ecosystems. A key componen
...
+
+The quickest way is to use docker, even without cloning the repo.  Run this command and open your browser to http://localhost:8888 .  Ctrl+C to stop the server.
```
pathmapper: In my opinion the quickest way is to use the online editor and using the CLI is also fast. So I think we don't need to say which is the quickest way. Not every user is familiar with docker.

orangemug:
Thanks @nyurik. A quick question, am I correct in thinking that the 2 stage build means that `node` won't be available in the final image? If so I think we'd want to make a `Dockerfile` and `Dockerfile.prod` (or something like that) because otherwise we won't be able to do development with this image. Currently I use https://github.com/orangemug/maputnik-services for all my development, however any development will probably require an `npm install`.

Let me know your thoughts 

Note: I'm away on holiday at the moment so can't try it out, back next week.

package.json
```diff
@@ -17,6 +17,9 @@
     "url": "https://github.com/maputnik/editor"
   },
   "author": "Lukas Martinelli",
+  "contributors": [
+    "Yuri Astrakhan <YuriAstrakhan@gmail.com>"
```
orangemug: Can you remove this. Reasoning, I don't want to start adding all the [contributors](https://github.com/maputnik/editor/graphs/contributors) to the `package.json` when it's already in [GitHub](https://github.com/maputnik/editor/graphs/contributors)

README.md
nyurik: Makes sense about grouping them together.

README.md
nyurik: > Thanks @nyurik. A quick question, am I correct in thinking that the 2 stage build means that `node` won't be available in the final image? If so I think we'd want to make a `Dockerfile` and `Dockerfile.prod` (or something like that) because otherwise we won't be able to do development with this image. Currently I use https://github.com/orangemug/maputnik-services for all my development, however any development will probably require an `npm install`.

@orangemug could you describe your dev process?  We could create `Dockerfile.dev` (I would prefer to make .dev a special case, because most of the time Dockerfile is built for production), but we have to be clear about the dev pattern here -- e.g. share project root dir with docker, and let docker create node_modules (might be a bit painful with the correct file ownership) and have a long docker command + options to start it right.  Or we could make docker.dev that is fully node-based, and uses `npm start` as its CMD.

nyurik: Clarification: essentially my confusion is how you build:
* if you build locally, you don't need docker
* if you build in docker, you either copy the entire dir into the docker, in which case it doesn't really matter if it has node or not
* or, you share source files with the docker, in which case node_modules are still local, and docker may mess up the file permissions unless you take extra care to run it as local user.

orangemug: Hey @nyurik all my nodejs development happens in containers, really for the ease of setup between environments, and so I never have to worry about installing multiple versions of nodejs etc... For simplier non-maputnik editor projects I use https://github.com/orangemug/nde, including modules I write, for example https://github.com/mgljs-contrib/icepick-style.

For the maputnik editor I use <https://github.com/orangemug/maputnik-services>, basically it'll give me a full maputnik setup website and all, from a single command `docker-compose up` it mounts the code volume, so I edit all my code locally on my machine. If I need to `npm install` that happens via a `docker-compose exec editor npm install uuid`. It seems 
a pretty robust approach, I've been using this pretty much since I started development with Maputnik. I also use a simpler approach in my day job(s) the the last couple of years.

For the sake of this pull request as long as there is a `Dockerfile.dev` which has `node`/`npm` as well as the other common deps (`git`, `python`) available I'm happy. I can alter <https://github.com/orangemug/maputnik-services> to use that file instead. Happy for you just to just copy over the current `Dockerfile` in `master` (`cp Dockerfile Dockerfile.dev`)

Thanks for the PR and let me know if you have any other questions

pathmapper:
@nyurik works great! Reduces the image size from 553 MB -> 211 MB.

nyurik:
I think I will actually switch to nginx -- far more stable, secure and performant for serving static content, and it is about 60MB smaller.

README.md
nyurik: @orangemug the [node image](https://hub.docker.com/_/node) has python and git built-in. The current Dockerfile uses the `node-slim` and adds those manually, but with the new dockerfile, there is no longer a need to use `node-slim` as the base. Can you just use the standard `node` image in your docker-compose?   If you need to run a command, you can just [run it directly](https://stackoverflow.com/questions/30063907/using-docker-compose-how-to-execute-multiple-commands) from the docker-compose file.  And you clearly don't need the dockerfile's `copy` command because your code stays local -- so why have a separate dockerfile?

orangemug:
> I think I will actually switch to nginx -- far more stable, secure and performant for serving static content, and it is about 60MB smaller.

@nyurik did you want to add nginx? You could also use https://hub.docker.com/_/haproxy and save a whole extra MB over nginx 😁 (see https://github.com/orangemug/maputnik-services/blob/master/config/haproxy/haproxy.cfg for config)

Although honestly I'm happy with the PR as is, we can always optimise later. It's up to you, if you're happy I'll merge it, else I'll hold off for a bit.

Note: We're adding CI via GitHub actions in this PR https://github.com/maputnik/editor/pull/687 which deploys the docker image to the GitHub registry. There's still a bit of work to do on deployment it currently only publishes as 'master' whereas I guess we'll want it to deploy the tags as well.

nyurik:
I'm all for iterative steps, so if you +1 it, i will merge it.

orangemug:
Thanks, merged!

Issue #502, opened 2019-03-10
JSON editor: Cursor jumps to the end of the JSON code 


pathmapper:
As reported by @zstadler and @sk1p in https://github.com/maputnik/editor/issues/484, for certain inputs the cursor jumps to the end of the JSON code.

This bug exists since `v1.1.0` (`v1.0.2` was working fine).

Perhaps it was introduced with the switch from `react-codemirror` to `react-codemirror2 ` in https://github.com/maputnik/editor/commit/934a994ac50af66e5b7d210612b9e639fdbfca50.

pathmapper:
Steps to reproduce:
1. Open _OSM Liberty_ style
2. Select `poi_z16` layer
3. In _JSON Editor_ delete the first value (which is `0`) of the `text-offset` property
4. Enter `-` and `0`

![cursor_jumps_to_end](https://user-images.githubusercontent.com/20856381/66766818-55578f80-eeaf-11e9-88d8-10b4d7a93b60.gif)

pathmapper:
@zstadler and @sk1p 
While working on performance improvements in https://github.com/maputnik/editor/pull/562, @orangemug removed `react-codemirror2` which seems to fix this cursor issue.

Maybe you want to give it a try:
https://1527-84182601-gh.circle-artifacts.com/0/artifacts/build/index.html

pathmapper:
Thanks for the feedback @sk1p, #562 is merged now.

PR #557, opened 2019-10-13
LayerList performance improvements

orangemug:
So to follow on from @pathmapper comment in https://github.com/maputnik/editor/pull/521#issuecomment-493949882 it does indeed look like the performance has suffered between `v1.5.0` and `v1.6.0`. Although the 84ms mentioned in https://github.com/maputnik/editor/issues/372 was clearly just incorrect results, see results from v1.5.0 below (no idea what mistake I made perviously)

The condition I'm testing is changing the background color of osm-liberty. The reults from the devtools performance profiler are as follows

v1.5.0

Screen Shot 2019-10-13 at 11 31 26

v1.6.0

Screen Shot 2019-10-13 at 11 30 11

If I run it in development it's significantly slower, I'm assuming all the react profiling metrics is causing some of that slow down. But for the following demonstration, lower numbers are better and don't compare them to the production builds above.

So the current v1.6.0 branch run in development is taking around 700-800ms to render a frame once the background color is changed.

Screen Shot 2019-10-13 at 11 12 52

However if we dive a little deeper in the call trace we can see that the sortable list is causing most of the issues. 

Screen Shot 2019-10-13 at 11 12 59

This is wasted CPU cycles because rather than rendering every component to work out what has changed. We can instead just check the layers list for `id` or `layout.visibility` changes, because thats the only thing that's used in sub components of `<LayerList/>`. This PR adds a   `shouldComponentUpdate` function to check for that condition.

With that update applied, I rebuilt a production build of the app with those changes and retested.

Demo: https://1505-84182601-gh.circle-artifacts.com/0/artifacts/build/index.html

Screen Shot 2019-10-13 at 11 33 18

Now we're back down to ~140ms for a frame, which is slightly better than v1.5.0, and significantly better than v1.6.0. 

For anybody outside the Maputnik team reading this, we know that 140ms isn't ideal performance. We do have further improvements in the works, see #521, however those will take a little more time to get right.

I also throttled `<ChromePicker/>` to 30fps, I didn't notice any major improvement with this applied. However it's good practice to throttle updates.

All the testing was completed on the same laptop, within a short time window with no other apps running except chrome, a terminal and the activity monitor.

### DIFF
--- a/config/webpack.production.config.js
+++ b/config/webpack.production.config.js
@@ -18,6 +18,9 @@ module.exports = {
     filename: '[name].[contenthash].js',
     chunkFilename: '[contenthash].js'
   },
+  optimization:{
+    minimize: false,
+  },
   resolve: {
     extensions: ['.js', '.jsx']
   },
--- a/config/webpack.production.config.js
+++ b/config/webpack.production.config.js
@@ -18,9 +18,6 @@ module.exports = {
     filename: '[name].[contenthash].js',
     chunkFilename: '[contenthash].js'
   },
-  optimization:{
-    minimize: false,
-  },
   resolve: {
     extensions: ['.js', '.jsx']
   },
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
     "detect-browser": "^4.5.0",
     "file-saver": "^1.3.8",
     "jsonlint": "github:josdejong/jsonlint#85a19d7",
+    "lodash": "^4.17.15",
     "lodash.capitalize": "^4.2.1",
     "lodash.clamp": "^4.0.3",
     "lodash.clonedeep": "^4.5.0",
--- a/src/components/fields/ColorField.jsx
+++ b/src/components/fields/ColorField.jsx
@@ -30,7 +30,7 @@ class ColorField extends React.Component {
   }
 
   onChangeNoCheck (v) {
-    this.props.onChange(formatColor(v));
+    this.props.onChange(v);
   }
 
   //TODO: I much rather would do this with absolute positioning
--- a/src/components/fields/ColorField.jsx
+++ b/src/components/fields/ColorField.jsx
@@ -2,6 +2,7 @@ import React from 'react'
 import Color from 'color'
 import ChromePicker from 'react-color/lib/components/chrome/Chrome'
 import PropTypes from 'prop-types'
+import lodash from 'lodash';
 
 function formatColor(color) {
   const rgb = color.rgb
@@ -21,6 +22,15 @@ class ColorField extends React.Component {
 
   state = {
     pickerOpened: false
+  }
+
+  constructor () {
+    super();
+    this.onChangeNoCheck = lodash.throttle(this.onChangeNoCheck, 1000/30);
+  }
+
+  onChangeNoCheck (v) {
+    this.props.onChange(formatColor(v));
   }
 
   //TODO: I much rather would do this with absolute positioning
@@ -82,7 +92,7 @@ class ColorField extends React.Component {
       }}>
       <ChromePicker
         color={currentColor}
-        onChange={c => this.props.onChange(formatColor(c))}
+        onChange={c => this.onChangeNoCheck(formatColor(c))}
       />
       <div
         className="maputnik-color-picker-offset"
--- a/src/components/layers/LayerList.jsx
+++ b/src/components/layers/LayerList.jsx
@@ -1,6 +1,7 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 import classnames from 'classnames'
+import lodash from 'lodash';
 
 import LayerListGroup from './LayerListGroup'
 import LayerListItem from './LayerListItem'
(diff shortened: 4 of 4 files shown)


pathmapper:
Hmm, do you get consistent values for multiple test runs @orangemug ?

Asking because I run multiple test with production builds of `v1.6.0`, `v.1.6.1` and your `demo build` and can't say that one of these gives better results than the others.

Used the latest versions of Firefox and Chome (with Firefox the range of the results is wider).

pathmapper:
Here are the results of 10 test runs for each production build. All on Chrome, all the same test setup.
Test: Changing the background color of osm-liberty using the color picker (click)

|Test run | ms (PR demo) | ms (v1.6.1) | ms (v1.6.0) |
| :------------- | -------------: | -------------: | -------------: |
| 1 |  100.1 | 115.8 | 103.0
| 2|  106.2 | 61.9| 101.4
| 3 |  83.7 | 97.4| 94.0
|  4|   100.4| 91.4| 94.1
|  5|  110.7 | 148.2| 77.3
|  6|   58.8| 92.9| 92.6
|  7|   63.9| 74.5| 81.4
|  8|   85.2| 53.8| 96.5
|  9|  79.3 | 85.7| 86.5
|  10|   78.0| 82.5| 90.4

orangemug:
Ok so here's where I'm at with this PR. So my initial condition wasn't very good (sorry) background color will trigger a color transition in mapbox-gl-js. So depending on what color you pick you get different performance from mapbox-gl-js, which will in turn effect frame rate. Ideally we want to just test the react/UI bit, the stuff we have control over.  

So I've changed my condition for the performance metrics I'm now changing the 'max zoom' of the 'state' layer osm-liberty. In a way that it would cause a re-render of the mapbox-gl map. So for example with the current map zoom at 1, I would change the 'max zoom' of the 'state' layer between `6` and `7`

I've also install the react chrome extension, so I can do profiling in that instead in development. Here are my development results

Without `shouldComponentUpdate`

Screen Shot 2019-10-14 at 10 09 42
Screen Shot 2019-10-14 at 10 09 21
Screen Shot 2019-10-14 at 10 08 53

With `shouldComponentUpdate`
    
Screen Shot 2019-10-14 at 10 11 19
Screen Shot 2019-10-14 at 10 11 02
Screen Shot 2019-10-14 at 10 10 26

I then built the 2 different versions and tested with the regular chrome devtools performance tab. I tested 8x3 times to reduce the effect of the CPU being busy with background tasks.  

```
irb(main):004:0> (85.0+74+82+110+124+91+94+109)/8 # new code
=> 96.125
irb(main):001:0> (108.0+82+97+78+135+149+156+111)/8  #1.6.1
=> 114.5
irb(main):006:0> (99.0+134+108+110+115+100+83+158)/8 # new code
=> 113.375
irb(main):008:0> (141.0+136+130+95+144+130+116+97)/8 # 1.6.1
=> 123.625
irb(main):009:0> (81.0+71+95+74+75+86+78+88)/8  # new code
=> 81.0
irb(main):010:0> (137.0+135+103+104+116+110+125+97)/8  # 1.6.1
=> 115.875 
```

So if we average across the test runs for the new code and 1.6.1 I get the following results

```
irb(main):002:0> (96.125+113.375+81.0)/3  # new code
=> 96.83333333333333
irb(main):003:0> (114.5+123.625+115.875)/3  # 1.6.1
=> 118.0
```

Hardly amazing results but 20ms is better than nothing.

Next steps

 - I think I'll add a [react profiling-in-production](https://gist.github.com/bvaughn/25e6233aeb1b4f0cdb8d8366e54a3977#profiling-in-production) as a separate build in our CI, so we can profile easier
 - I also noticed from the profiling that the JSONEditor is effecting performance in a big way (more investigation required)

pathmapper:
Thanks for the detailed explanation @orangemug . With the `max-zoom` test I'm seeing also a performance improvement. :tada: 

> 20ms is better than nothing

 18% increased performance sounds better :-)

> the JSONEditor is effecting performance in a big way (more investigation required)

Perhaps this annoying bug is related: https://github.com/maputnik/editor/issues/502

pathmapper:
> I think I'll add a react profiling-in-production as a separate build in our CI, so we can profile easier

The intention is that we could use the react chrome extension on this CI build?

orangemug:
> The intention is that we could use the react chrome extension on this CI build?

Yeah that's right, so what I'm thinking is that we'll create 2 artifacts from the CI builds one will be the normal build we use to show in demos currently. The other for performance testing we'll include the [react profiling-in-production](https://gist.github.com/bvaughn/25e6233aeb1b4f0cdb8d8366e54a3977#profiling-in-production), also add a debug option to maputnik were we can disable the renderer (mapbox-gl, ol) then we can get good metrics about the performance of our UI in isolation. 

> 18% increased performance sounds better :-)

👍 

> Perhaps this annoying bug is related: #502

I'll dig into the JSONEditor performance in another ticket.

Thanks again for all the help with this PR @pathmapper, having to justify my metrics really helped. I think we have a better approach moving forward.

Merging!

PR #561, opened 2019-10-15
Set a color for the OpenLayers attribution control

fredj:
The color is currently white on a white div:
Screenshot from 2019-10-15 14-35-17


orangemug:
Thanks @fredj merging!

PR #562, opened 2019-10-15
react-codemirror2 -> codemirror directly

orangemug:
I've changed the `<JSONEditor/>` from [react-codemirror2](https://www.npmjs.com/package/react-codemirror2) to interacting with [codemirror](https://codemirror.net/) directly. Some initial metrics

With the same test case, as described in https://github.com/maputnik/editor/pull/557#issuecomment-541628824

> So I've changed my condition for the performance metrics I'm now changing the 'max zoom' of the 'state' layer osm-liberty. In a way that it would cause a re-render of the mapbox-gl map. So for example with the current map zoom at 1, I would change the 'max zoom' of the 'state' layer between 6 and 7

Before change

Screen Shot 2019-10-15 at 17 45 07

After change

Screen Shot 2019-10-15 at 17 43 46

Also if you just grab the color picker for the background layer and hold it while moving you mouse to continually change color you'll notice it's much smoother. Although not a very 'pure' test, you can feel the difference. You can also see many less 'long frame' warnings in chrome and lower frame times

Before

Screen Shot 2019-10-15 at 17 59 47
Screen Shot 2019-10-15 at 18 00 05

After

Screen Shot 2019-10-15 at 18 00 20
Screen Shot 2019-10-15 at 18 00 30

This still requires further testing, if anybody wants to help me find bugs the build URL is https://1527-84182601-gh.circle-artifacts.com/0/artifacts/build/index.html

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9069,11 +9069,6 @@
         "prop-types": "^15.5.10"
       }
     },
-    "react-codemirror2": {
-      "version": "6.0.0",
-      "resolved": "https://registry.npmjs.org/react-codemirror2/-/react-codemirror2-6.0.0.tgz",
-      "integrity": "sha512-D7y9qZ05FbUh9blqECaJMdDwKluQiO3A9xB+fssd5jKM7YAXucRuEOlX32mJQumUvHUkHRHqXIPBjm6g0FW0Ag=="
-    },
     "react-collapse": {
       "version": "4.0.3",
       "resolved": "https://registry.npmjs.org/react-collapse/-/react-collapse-4.0.3.tgz",
--- a/package.json
+++ b/package.json
@@ -47,7 +47,6 @@
     "react-aria-modal": "^4.0.0",
     "react-autobind": "^1.0.6",
     "react-autocomplete": "^1.8.1",
-    "react-codemirror2": "^6.0.0",
     "react-collapse": "^4.0.3",
     "react-color": "^2.17.3",
     "react-dom": "^16.10.2",
--- a/src/components/layers/JSONEditor.jsx
+++ b/src/components/layers/JSONEditor.jsx
@@ -1,9 +1,9 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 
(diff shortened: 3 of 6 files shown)


pathmapper:
:rocket: 

Some first comments:
 - `react-codemirror2` could be removed from `package.json`
https://github.com/maputnik/editor/blob/master/package.json#L50
- this should be adapted: https://github.com/maputnik/editor/blob/master/src/styles/_react-codemirror.scss
- the JSON editor should use the full available width of its window
- https://github.com/maputnik/editor/issues/502 seems to be fixed, at least the test case in https://github.com/maputnik/editor/issues/502#issuecomment-541778685

orangemug:
> react-codemirror2 could be removed from package.json
https://github.com/maputnik/editor/blob/master/package.json#L50

Fixed in https://github.com/maputnik/editor/pull/562/commits/ef23f01e676f74bbac691ba76a9405c8c353040f

> this should be adapted: https://github.com/maputnik/editor/blob/master/src/styles/_react-codemirror.scss

Good spot fixed in https://github.com/maputnik/editor/pull/562/commits/22b6a4a2bfec1985ceee1e262dd95e34439bb13d

> #502 seems to be fixed, at least the test case in #502 (comment)

Yeah agree seems to be fixed.

pathmapper:
Sometimes after an input the scroll position of the JSON editor is changed when it's not needed. Could this be avoided?

Last step when inserting `,`:

Style: _OSM Liberty_, Layer: _road_link_casing_
scroll_position

orangemug:
> Sometimes after an input the scroll position of the JSON editor is changed when it's not needed. Could this be avoided?

Hmm, looks like that's an issue with the exiting editor also (see https://maputnik.github.io/editor/) so probably a codemirror thing. Suggest we raise a new issue and look into this separately as it may take a while to find a solution.

pathmapper:
Agreed @orangemug, tracked in https://github.com/maputnik/editor/issues/565.

orangemug:
@pathmapper are we happy to merge this PR now?

pathmapper:
@orangemug yep, merging!

Issue #560, opened 2019-10-14
Setting root "center" and "zoom"


fredj:
Hi all,

Is there a way to set the [center](https://docs.mapbox.com/mapbox-gl-js/style-spec/#root-center) and the [zoom](https://docs.mapbox.com/mapbox-gl-js/style-spec/#root-zoom) of a style?

If it's not already the case, it would be nice to add them in the "Export style" form, with the default values from the url

pathmapper:
Hi @fredj,

> Is there a way to set the center and the zoom of a style?

No, that's currently not possible, thanks for the suggestion.

Issue #440, opened 2018-10-22
Add support for global light source


orangemug:
This should be added to the settings modal with the other global style settings.

See <https://www.mapbox.com/mapbox-gl-js/style-spec/#light>

pathmapper:
Added by @orangemug in https://github.com/maputnik/editor/pull/563

PR #563, opened 2019-10-16
Added support for more root level properties

orangemug:
@fredj @pathmapper as a fix for #560 how about adding them to the style settings? 

Screen Shot 2019-10-16 at 07 44 48

Demo: https://1538-84182601-gh.circle-artifacts.com/0/artifacts/build/index.html

### DIFF
--- a/src/components/fields/PropertyGroup.jsx
+++ b/src/components/fields/PropertyGroup.jsx
@@ -59,7 +59,7 @@ export default class PropertyGroup extends React.Component {
         onChange={this.onPropertyChange}
         key={fieldName}
         fieldName={fieldName}
-        value={fieldValue === undefined ? fieldSpec.default : fieldValue}
+        value={fieldValue}
         fieldSpec={fieldSpec}
       />
     })
--- a/src/components/inputs/ArrayInput.jsx
+++ b/src/components/inputs/ArrayInput.jsx
@@ -17,27 +17,84 @@ class ArrayInput extends React.Component {
     default: [],
   }
 
+  constructor (props) {
+    super(props);
+    this.state = {
+      value: this.props.value.slice(0),
+      // This is so we can compare changes in getDerivedStateFromProps
+      initialPropsValue: this.props.value.slice(0),
+    };
+  }
+
+  static getDerivedStateFromProps(props, state) {
+    const value = [];
+    const initialPropsValue = state.initialPropsValue.slice(0);
+
+    Array(props.length).fill(null).map((_, i) => {
+      if (props.value[i] === state.initialPropsValue[i]) {
+        value[i] = state.value[i];
+      }
+      else {
+        value[i] = state.value[i];
+        initialPropsValue[i] = state.value[i];
+      }
+    })
+
+    return {
+      value,
+      initialPropsValue,
+    };
+  }
+
+  isComplete (value) {
+    return Array(this.props.length).fill(null).every((_, i) => {
+      const val = value[i]
+      return !(val === undefined || val === "");
+    });
+  }
+
   changeValue(idx, newValue) {
-    console.log(idx, newValue)
-    const values = this.props.value.slice(0)
-    values[idx] = newValue
-    this.props.onChange(values)
+    const value = this.state.value.slice(0);
+    value[idx] = newValue;
+
+    this.setState({
+      value,
+    }, () => {
+      if (this.isComplete(value)) {
+        this.props.onChange(value);
+      }
+      else {
+        // Unset until complete
+        this.props.onChange(undefined);
+      }
+    });
   }
 
   render() {
+    const {value} = this.state;
+
+    const containsValues = (
+      value.length > 0 &&
+      !value.every(val => {
+        return (val === "" || val === undefined)
+      })
+    );
+
     const inputs = Array(this.props.length).fill(null).map((_, i) => {
       if(this.props.type === 'number') {
         return <NumberInput
           key={i}
-          default={this.props.default[i]}
-          value={this.props.value[i]}
+          default={containsValues ? undefined : this.props.default[i]}
+          value={value[i]}
+          required={containsValues ? true : false}
           onChange={this.changeValue.bind(this, i)}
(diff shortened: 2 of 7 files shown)


fredj:
Very nice, thanks @orangemug !

May I suggest to keep the "Style Renderer" option at the bottom of the list?

orangemug:
@fredj @pathmapper also added support for the root level 'light' properties

Screen Shot 2019-10-16 at 08 27 56

orangemug:
> May I suggest to keep the "Style Renderer" option at the bottom of the list?

Done see screenshot above

pathmapper:
@orangemug great, closes also https://github.com/maputnik/editor/issues/440.

- For `Center` maybe add labeling which indicates which field is `lon` and which `lat` ?
- For `Light Position`  maybe add labelling `r`, `a`, `p` to make clear which field belongs to which value?

I'll do some more testing.

pathmapper:
Some more findings and thoughts:

- `Light anchor` defaults to `viewport` (currently `map` is highlighted as default)
- Add also `transition`: https://docs.mapbox.com/mapbox-gl-js/style-spec/#root-transition
- When the `Style Settings` modal is open, the map is darkened a bit, which alters the effect of the light source. Not sure if we want to change this, but currently there is a need to close the `Style Settings` modal to see the _real_ effect.  
- Not needed for this PR but just wanted to mention it: A nice feature would be a button in the `Settings Modal`, which populates the values for `center`, `zoom`, `pitch` and `bearing` with the values from the URL hash.

fredj:
> Light anchor defaults to viewport (currently map is highlighted as default)

It's already the default value but it looks like the `default` property is not used in the [EnumInput](https://github.com/maputnik/editor/blob/95e205943af1653f98e72c51e34a8b786cd0004e/src/components/inputs/EnumInput.jsx#L21) component

orangemug:
> Add also transition: https://docs.mapbox.com/mapbox-gl-js/style-spec/#root-transition

Fixed in https://github.com/maputnik/editor/pull/563/commits/d0a47bd1226f4c7099f7d97fd374b6902b9381d4

> `Light anchor` defaults to `viewport` (currently `map` is highlighted as default)

Fixed in https://github.com/maputnik/editor/pull/563/commits/8c760bb8108dbc8991f2d918f5d920ca7c0726d0

Found another bug: You also can't remove a property once set, for example set 'Light intensity' then try and remove it. I'll have a go at fixing that later.

Thanks for all the help @fredj @pathmapper

orangemug:
New build: https://1565-84182601-gh.circle-artifacts.com/0/artifacts/build/index.html

pathmapper:
The `Light position` and `Style Renderer` tooltips are not fully visible and when scrolled down, the scroll position is set back to the initial position.

tooltip_issue

orangemug:
> This issue exists also for some tooltips of the Data Sources modal.

If you collapse the JSON editor on a layer editing panel it's also an issue there as well. I think we should either find a good react tooltip library or alternatively a different UI for help text. I might be better to, on hover show a `?` next to the label and then on click show link to the place in the mapbox-gl docs (or our own render of the mapbox-gl docs)

pathmapper:
All right, moved this to https://github.com/maputnik/editor/issues/567

fredj:
I really hope that the ability to set the `center` property will come back !

orangemug:
Yeah sorry, I noticed it was actually a little broken last night, so remove it temporarily. Yeah I plan to work on it next time a get the chance. Sorry I didn't have a chance to leave a comment last night @fredj.

A little description of the issue, so the `<ArrayInput />` component allows you to enter a number of values as a single property, so for example a coordinate `[x,y]` or `[x,y,z]`, unfortunately that `<ArrayInput />` doesn't really work because if you enter a value in `x` then move focus to `y` you    end up triggering a change event. That in turn does validation on the value `[x,y,z]` (via mapbox-gl spec validator), which is invalid because you haven't yet entered anything in `y` or `z`.

Note this kind of worked for the `center` because if you entered a value in `x` (for example `1.1` the value would be `[1.1]`, the `center` validation passes for both `[x]` and `[x,y]`. However if you enter a value in `y` first the value would be `[undefined, 1.1]` which would fail validation

The 'light position' didn't work at all because the validation asserts that the length must be `3` so you enter a value in `x` it'll fail validation right away.

The solution I'm kind of leaning towards is making a 'coordinate' input type only triggering the `onChange` event when all the inputs contain a value.

Any suggestions or comments welcome.

fredj:
Thanks for the explanations and your work on this pull request @orangemug

fredj:
> Any suggestions or comments welcome.

Could it be implemented using the same `state.editing` / `getDerivedStateFromProps` technique used in `StringInput`?
And `editing` is true if `this.props.value` length is the same as `this.props.length` (and without `undefined` value).

(I've a very limited experience in react, so it's probably a wrong idea!)

orangemug:
So I've implemented an idea of what I think `<ArrayInput />` should look like, based off what @fredj commented. So what it now does is it doesn't trigger `onChange` with a value until all the inputs are filled with values. If you fill `x` but not `y` the `onChange`event will trigger with `onChange(undefined)` effectively removing the value from the JSON. 

I've also made fixes to the array input to allow for default values and removing of values, which fixes bugs with `*-translate`

Demo: https://1592-84182601-gh.circle-artifacts.com/0/artifacts/build/index.html

Let me know what you think @pathmapper @fredj

pathmapper:
@orangemug thanks, works great.

If you remove the values for `light` (so that the default values are used), `"light": {}` remains in the style which is not needed. Same for `transition`. But it could also stay like it is now.

fredj:
looks very good, thanks again @orangemug

orangemug:
> If you remove the custom values for light (so that the default values are used), "light": {} remains in the style which is not needed. Same for transition. But the PR could also be merged like it is now.

Nice spot, lets merge this now, I'll fix that in another PR.

Thanks for all the help @fredj @pathmapper, merging!

PR #566, opened 2019-10-16
Added profiling build to CI

orangemug:
From https://github.com/maputnik/editor/pull/557#issuecomment-542103025

> Yeah that's right, so what I'm thinking is that we'll create 2 artifacts from the CI builds one will be the normal build we use to show in demos currently. The other for performance testing we'll include the react [profiling-in-production](https://gist.github.com/bvaughn/25e6233aeb1b4f0cdb8d8366e54a3977#profiling-in-production)

This PR adds a new build with profiling enabled to the artifacts directory `/profiling` so we can use the react profiler

Demo https://1558-84182601-gh.circle-artifacts.com/0/artifacts/profiling/index.html

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,7 +18,6 @@ templates:
 
     - run: mkdir -p /tmp/artifacts/logs
     - run: npm run build
-    - run: npm run profiling-build
     - run: npm run lint
     - run: npm run lint-styles
     - store_artifacts:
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -18,6 +18,7 @@ templates:
 
     - run: mkdir -p /tmp/artifacts/logs
     - run: npm run build
+    - run: npm run profiling-build
```
pathmapper: Like this there is no profiling-build for linux-node-v10 which runs under `&wdio-steps`.

Do we need a profiling build for all node versions and for linux and osx? Or would it be enough to add it only to 
`&wdio-steps` to have a profiling build for linux node LTS like the wdio tests?

image

orangemug: Good idea @pathmapper I've updated the PR

pathmapper:
Thanks @orangemug, works great!

PR #569, opened 2019-10-16
Fix the layer type test in LayerEditor component

### LINKED ISSUE #568: "Source Layer" input is always displayed in the layer editor
The input must be displayed only for vector tile sources but it's currently always here (for background, heatmap, ...)

Screenshot from 2019-10-16 15-27-00

fredj:
`LayerSourceLayerBlock` was always included

fixes #568


orangemug:
Nice find, merging!

PR #575, opened 2019-10-24
set OS Zoomstack endpoints to v2

snodnipper:
Update the OS Zoomstack endpoints to the second release


orangemug:
Looks good, thanks @snodnipper, merging!

PR #579, opened 2019-10-25
Function field fixes

orangemug:
Recent fixes to allowing you to reset the values of fields, caused issues with function fields. This was because previously those fields set the `value` of a property to it's `default`. Now we just pass the default value to the fields rather than mutating the value (see https://github.com/maputnik/editor/pull/563/files#diff-b98b1ae2788532d29e7217366330abaaL68). This resulted in the values being `undefined` when transitioning to 'function fields', causing the spec validator to throw an error.

This PR adds a new method `findDefaultFromSpec` which uses the default from the spec, else it falls back to a set of generic defaults (some properties don't have defaults).

I've also fixed `enum` to use the `<EnumInput />` again, which was previously defined in `<SelectInput/>`

### DIFF
--- a/src/components/fields/FunctionSpecField.jsx
+++ b/src/components/fields/FunctionSpecField.jsx
@@ -18,7 +18,7 @@ function isDataField(value) {
  * If we don't have a default value just make one up
  */
 function findDefaultFromSpec (spec) {
-  if (spec.default) {
+  if (spec.hasOwnProperty('default')) {
     return spec.default;
   }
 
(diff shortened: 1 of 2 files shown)


pathmapper:
I'm not sure which issue this PR fixes exactly, but there is still an issue, e.g. click on the _"Turn property into function ..."_ button of `Allow overlap`:

image

orangemug:
Nice find, fixed in latest commit

Demo: https://1731-84182601-gh.circle-artifacts.com/0/artifacts/build/index.html#0.35/0/0

pathmapper:
Thanks @orangemug, looks good now.

orangemug:
Thanks, merging

PR #582, opened 2019-10-27
Add open modal and source http/https errors

### LINKED ISSUE #374: Source without https doesn’t get warning
As one of the survey respondents pointed out, “source without https should get warning”.

As @orangemug pointed out, we can probably use https://www.mapbox.com/mapbox-gl-js/api/#map.event:error here and print any errors that occur when loading resources.

orangemug:
If the editor is loaded from `https://` (like on `https://maputnik.github.io/editor/`) show an error for url fields where the user is trying to use `http://` urls 

Demo: <https://1815-84182601-gh.circle-artifacts.com/0/artifacts/build/index.html>

Screen Shot 2019-10-27 at 17 47 54
Screen Shot 2019-10-27 at 17 48 12

Fixes #374

### DIFF
--- a/src/components/inputs/UrlInput.jsx
+++ b/src/components/inputs/UrlInput.jsx
@@ -42,6 +42,10 @@ class UrlInput extends React.Component {
     required: PropTypes.bool,
   }
 
+  static defaultProps = {
+    onInput: () => {},
+  }
+
   constructor (props) {
     super(props);
     this.state = {
@@ -53,6 +57,7 @@ class UrlInput extends React.Component {
     this.setState({
       error: validate(url)
     });
+    this.props.onInput(url);
   }
 
   render () {
--- a/src/components/inputs/UrlInput.jsx
+++ b/src/components/inputs/UrlInput.jsx
@@ -3,6 +3,34 @@ import PropTypes from 'prop-types'
 import StringInput from './StringInput'
 import SmallError from '../util/SmallError'
 
+
+function validate (url) {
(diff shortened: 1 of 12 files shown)


pathmapper:
Very nice @orangemug, should this work also for 
- Sprite URL (_Style Settings_)
- Glyphs URL (_Style Settings_)
- GeoJSON (URL) (_Data Sources_) ?

orangemug:
Yeah @pathmapper you're right, added that in, demo: https://1854-84182601-gh.circle-artifacts.com/0/artifacts/build/index.html#0.35/0/0

src/components/inputs/UrlInput.jsx
```diff
@@ -0,0 +1,77 @@
...
+  };
+  const protocol = getProtocol(url);
+  if (
+    protocol &&
+    protocol === "http:" &&
+    window.location.protocol === "https:"
+  ) {
+    error = (
+      <SmallError>
+        CORs policy won&apos;t allow fetching resources served over http from https, use a <code>https://</code> domain
```
pathmapper:
```suggestion
        CORS policy won&apos;t allow fetching resources served over http from https, use a <code>https://</code> domain
```

Or is the lower case `s` intentional?

orangemug: Nope you're completely right

pathmapper:
Looks good! I like how the default protocol is set for the example URLs, very nice enhancement,

PR #602, opened 2020-01-18
Update react-collapse

### LINKED ISSUE #565: JSON editor: unwanted scroll position change
Sometimes after an input the scroll position of the JSON editor is changed when it's not needed.

Last step when inserting `,`:

Style: _OSM Liberty_, Layer: _road_link_casing_
scroll_position

orangemug:
I believe this fixes #565 

`react-collapse` was updating the DOM resetting the scroll position of the collapsible element, which was causing the scroll position jump in the pane. The new version of `react-collapse` seems to fix this issue.

There is a demo ready for testing: <https://1910-84182601-gh.circle-artifacts.com/0/artifacts/build/index.html>

### DIFF
--- a/src/components/layers/JSONEditor.jsx
+++ b/src/components/layers/JSONEditor.jsx
@@ -105,17 +105,6 @@ class JSONEditor extends React.Component {
   }
 
   render() {
-    const codeMirrorOptions = {
-      mode: {name: "javascript", json: true},
-      tabSize: 2,
-      theme: 'maputnik',
-      viewportMargin: Infinity,
-      lineNumbers: true,
-      lint: true,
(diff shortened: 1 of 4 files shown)


src/components/layers/JSONEditor.jsx
```diff
@@ -105,17 +105,6 @@ class JSONEditor extends React.Component {
   }
 
   render() {
-    const codeMirrorOptions = {
```
orangemug: This was just an old object that was no longer used.

pathmapper:
Thanks @orangemug, yes it seems like this fixes #565, the issue is not reproducible anymore with the test case.

Issue #427, opened 2018-10-20
Emulation mode for SDKs (JS/iOS/Android) support


gregwolanski:

orangemug:
So I guess another way to look at this would be to have a 'emulation mode' and filter out fields that are not supported on certain platforms. So if 'iOS emulation' was selected a number of fields would be disabled and marked as so in the editor. 

My thinking is that listing what is supported might not be that helpful, because ultimately you want to target a style at a set of platforms. So if a feature is missing from a platform you want to see how that will effect your style.

Questions?

 - Does that make sense?
 - Is there something obvious I'm missing as to why that wouldn't work?
 - Any other thoughts?

orangemug:
> What happens if a style with unsupported features is used with iOS/Android/macOS? Are they
ignored or is the style not working in general?

That's my understanding, we'd need to check though

> It would be a lot of work to keep this all up-to-date with the style spec, in my opinion not possible to do it manually. Such emulation modes would need to be synced automatically with the style spec.

We have this information from the style-spec already, for example https://github.com/mapbox/mapbox-gl-js/blob/master/src/style-spec/reference/v8.json#L1019-L1025 

> There are also differences regarding what kind of features are supported within one platform, e.g. see SDK support for GeoJSON sources. So you would need a lot of different emulation modes.

True, I guess that's also true of the web target currently also though

pathmapper:
> True, I guess that's also true of the web target currently also though

Maybe it would be sufficiant to have an emulation mode only for the latest release of each platform.

pathmapper:
@orangemug this issue remains open because of the emulation mode, right?

orangemug:
Correct @pathmapper, although I don't think it's something we should prioritise anytime soon.

pathmapper:
Thanks @orangemug, updated the issue title.

pathmapper:
And yeah, I also think this is something for _later_.

Issue #425, opened 2018-10-20
Layout properties: Icons instead of text in “Cap” and “Join” properties


gregorywolanski:
![1](https://user-images.githubusercontent.com/787518/47254117-c364a480-d45d-11e8-81ba-c5bbe27300f1.png)

Opinions? :)

orangemug:
I'm not so sure I think this is a good idea. I think the text is probably more obvious than the icons would be. Questions

 - Does anybody else have any opinions?
 - What do other graphics/map editors do?

pathmapper:
Agree, let's close this.

It would be also difficult to have small icons which are showing the difference between `Butt` and `Square`.

The style spec has a text description of the different options: 
https://docs.mapbox.com/mapbox-gl-js/style-spec/#layout-line-line-cap
https://docs.mapbox.com/mapbox-gl-js/style-spec/#layout-line-line-join

Maybe we could make this description with tooltips available. Let's move this discussion to https://github.com/maputnik/editor/issues/567

PR #607, opened 2020-01-23
Inline property documentation and SDK docs

### LINKED ISSUE #567: Better solution for tooltips
From https://github.com/maputnik/editor/pull/563#issuecomment-542620356

> The Light position and Style Renderer tooltips are not fully visible and when scrolled down, the scroll position is set back to the initial position. This issue exists also for some tooltips of the Data Sources modal.

From https://github.com/maputnik/editor/pull/563#issuecomment-542644549

> If you collapse the JSON editor on a layer editing panel it's also an issue there as well. I think we should either find a good react tooltip library or alternatively a different UI for help text. I might be better to, on hover show a ? next to the label and then on click show link to the place in the mapbox-gl docs (or our own render of the mapbox-gl docs)

tooltip_issue

orangemug:
I've added inline documentation instead of tooltips as outlined in #567. I've also added SDK support to the property documentation where available. 

Fixes #567 (partial fix for #427)

Demo <https://2014-84182601-gh.circle-artifacts.com/0/artifacts/build/index.html>

Screenshot 2020-01-23 at 08 34 15


pathmapper:
Oh nice, just saw this PR after leaving comments in #425 and #567.

https://github.com/maputnik/editor/issues/567#issuecomment-577598795 is still valid, there's a part of the description missing for:
https://docs.mapbox.com/mapbox-gl-js/style-spec/#layout-line-line-cap
https://docs.mapbox.com/mapbox-gl-js/style-spec/#layout-line-line-join

Any chance to add it?

pathmapper:
These missing parts are nested in the style spec, e.g.:
https://github.com/mapbox/mapbox-gl-js/blob/master/src/style-spec/reference/v8.json#L783

orangemug:
Good spot @pathmapper, added in latest commit.

Demo <https://2020-84182601-gh.circle-artifacts.com/0/artifacts/build/index.html>

I've given it a quick test and it looks good, I'm a little short on time today though

orangemug:
New build with fixes https://2031-84182601-gh.circle-artifacts.com/0/artifacts/build/index.html

pathmapper:
I really like the inline documentation, some findings:

- `Sources` Modal:
When adding a source, the tooltip doesn't change when the source type is changed:

image


- There a no tooltips for the `Export` modal like they are for the token fields in the `Style Settings` modal.:

image

orangemug:
Thanks @pathmapper I've added the docs to the export modal and updated the 'type' in the sources modal 

With regards to the data source type, I've gone with a simple message instead of listing the possible values. As explained in a code comment (https://github.com/maputnik/editor/pull/607/files#diff-3648d61a30bd24d5ecf41c74cfb5fc05R170-R176) 

> Kind of a hack because the type changes, however maputnik has `1..n`
> options per type, for example 
>
>  - 'geojson' - 'GeoJSON (URL)' and 'GeoJSON (JSON)'
>  - 'raster' - 'Raster (TileJSON URL)' and 'Raster (XYZ URL)'
>
> So we just ignore the values entirely as they are self explanatory

New build <https://2035-84182601-gh.circle-artifacts.com/0/artifacts/build/index.html>

src/components/modals/SettingsModal.jsx
```diff
@@ -234,7 +244,7 @@ class SettingsModal extends React.Component {
         />
       </InputBlock>
 
-      <InputBlock label={"Style Renderer"} doc={"Choose the default Maputnik renderer for this style."}>
+      <InputBlock label={"Style Renderer"} fieldSpec={{doc: "Choose the default Maputnik renderer for this style."}}>
```
pathmapper: Maybe move this also to `fieldSpecAdditional`?

pathmapper:
> So we just ignore the values entirely as they are self explanatory

Good solution!

Sometimes there are URLs in the docs included, do we want them to be functional or leave them like they are now?

e.g. https://docs.mapbox.com/mapbox-gl-js/style-spec/root/#sprite

grafik

orangemug:
@pathmapper ah yeah forgot to mention that. So the docs are actually markdown, I think lets leave that for now. They are readable at the moment, I'll find a good small markdown parser (probably remark) and update in another PR.

pathmapper:
@orangemug sure, looks good to me now. One comment pending (https://github.com/maputnik/editor/pull/607#pullrequestreview-348644678), but this doesn't need to be addressed, just thought that you wanted to move all Maputnik docs there.

orangemug:
Thanks @pathmapper I've fixed the code review comment, merging!

PR #610, opened 2020-01-29
Fix for stuck/unchanging pop-up in Inspect view

### LINKED ISSUE #609: Error messages lead to stuck/unchanging pop-up in Inspect view
I just upgraded to the latest Maputnik (from 1.4, I think). Since the upgrade, I get a stream of errors in my browser console after enabling Inspect view and mousing over the map. Here's an example: 

```
react-dom.development.js:558 Warning: Encountered two children with the same key, `transportation_name-undefined`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.
    in div (created by FeaturePropertyPopup)
    in FeaturePropertyPopup
warningWithoutStack	@	react-dom.development.js:558
```

I get a lot of these errors mentioning several of my map data layers. After mousing around briefly, the same handful of features is displayed in the Inspect pop-up no matter where I go on the map. 

Anyone know how to fix this? 

In case it's relevant, I also get this error once: 

```
Warning: Failed prop type: Invalid prop `value` of type `boolean` supplied to `StringInput`, expected `string`.
    in StringInput (created by FeaturePropertyPopup)
    in FeaturePropertyPopup
printWarning | @ | checkPropTypes.js:20
```

orangemug:
Fixes #609

Demo <https://2043-84182601-gh.circle-artifacts.com/0/artifacts/build/index.html>

### DIFF
--- a/src/components/map/FeaturePropertyPopup.jsx
+++ b/src/components/map/FeaturePropertyPopup.jsx
@@ -21,8 +21,9 @@ function renderProperties(feature) {
   })
 }
 
-function renderFeature(feature) {
-  return <div key={`${feature.sourceLayer}-${feature.id}`}>
+function renderFeature(feature, idx) {
+  const uniqueKey = feature.hasOwnProperty('id') ? feature.id : idx;
+  return <div key={`${feature.sourceLayer}-${uniqueKey}`}>
     <div className="maputnik-popup-layer-id">{feature.layer['source-layer']}{feature.inspectModeCounter && <span> × {feature.inspectModeCounter}</span>}</div>
     <InputBlock key={"property-type"} label={"$type"}>
       <StringInput value={feature.geometry.type} style={{backgroundColor: 'transparent'}} />
--- a/src/components/map/FeaturePropertyPopup.jsx
+++ b/src/components/map/FeaturePropertyPopup.jsx
@@ -22,8 +22,7 @@ function renderProperties(feature) {
 }
 
 function renderFeature(feature, idx) {
-  const uniqueKey = feature.hasOwnProperty('id') ? feature.id : idx;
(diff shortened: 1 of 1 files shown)


src/components/map/FeaturePropertyPopup.jsx
```diff
@@ -21,8 +21,9 @@ function renderProperties(feature) {
   })
 }
 
-function renderFeature(feature) {
-  return <div key={`${feature.sourceLayer}-${feature.id}`}>
+function renderFeature(feature, idx) {
+  const uniqueKey = feature.hasOwnProperty('id') ? feature.id : idx;
```
orangemug: This is a little bit of a hack because ideally you shouldn't use an arrays index for uniqueness in react. However we don't have anything else unique to reference

See <https://github.com/mapbox/mapbox-gl-js/blob/b3046ea99a73cc027aa5d45ddba5285670097af8/src/util/vectortile_to_geojson.js#L8>

pathmapper: Just had a look at the spec:

https://github.com/mapbox/vector-tile-spec/tree/master/2.1#42-features

> A feature MAY contain an id field. If a feature has an id field, the value of the id SHOULD be unique among the features of the parent layer.

So if there are feature IDs and they are not unique (there is only _should_ in the spec) the issue would still exist, right?

What about using only the index here?

orangemug: Good spot @pathmapper

orangemug:
Changed to only using 'index' as per @pathmapper comment <https://github.com/maputnik/editor/pull/610#discussion_r372521304>

PR #613, opened 2020-02-02
Changed bracket matching color

orangemug:
This is so the bracket matching color doesn't clash with cursor color. I was becoming difficult for me to track my cursor as I moved the position over brackets. This would be particularly problematic for those with visual impairments.  

I've also moved all the codemirror css into a single place `./src/styles/_codemirror.scss`

Before
Screenshot 2020-02-02 at 11 25 38

After
Screenshot 2020-02-02 at 11 28 40


orangemug:
Yay, merging!

PR #614, opened 2020-02-02
Make background map container white to match html page default

orangemug:
HTML pages start with a white background in most browsers, most articles which contain maps are also going to be on a white background. So if a style doesn't set a background layer at the moment the user will probably get a different outcome than they expect when they export the style.

Before, with 'Empty Style'
Screenshot 2020-02-02 at 11 52 13

After, with 'Empty Style' 
Screenshot 2020-02-02 at 11 51 56

### DIFF
--- a/src/styles/_components.scss
+++ b/src/styles/_components.scss
@@ -1,5 +1,6 @@
 // MAP
 .maputnik-map__container {
+  background: white;
   display: flex;
   position: fixed !important;
   top: $toolbar-height + $toolbar-offset;


pathmapper:
Shall we also set the backgound color of the inspect mode to white?

https://github.com/maputnik/editor/blob/cc371d6a70dced526f7a31df5d2b4fb0c05df4b1/src/components/map/MapboxGlMap.jsx#L30

orangemug:
@pathmapper I think I like the black for inspect because if you start with an 'Empty style' and enter inspect mode you get a visual change even if no sources are added. Thoughts?

pathmapper:
Good point @orangemug, merging.

Issue #255, opened 2018-02-07
Url hash doesn't update when using the openlayers renderer


orangemug:
The hash in the url of the format `#0/0/0` doesn't update when changing map location/zoom while using the openlayers renderer

orangemug:
There are a few bugs and it needs a tidy but it's basically working  <https://github.com/orangemug/editor/compare/maintenance/openlayers-update...orangemug:fix/openlayers-hash>

orangemug:
The branch <https://github.com/orangemug/editor/compare/maintenance/openlayers-update...orangemug:fix/openlayers-hash> adds a fair bit more complexity. I'll leave this ticket open and that branch unmerged until we get better OpenLayers styling support.

pathmapper:
With zoom display UI and debug toolbox (https://github.com/maputnik/editor/pull/524) it's possible to display the current values of map `zoom`, `center` and `rotation`.

PR #617, opened 2020-02-17
Improved map control zoom styling

orangemug:
Also fixed zoom in/out button styling which was broken in previous `mapbox-gl-js` update.

Before
Screenshot 2020-02-17 at 15 45 04

After
Screenshot 2020-02-17 at 15 43 29


pathmapper:
Nice spot @orangemug - maybe use something different than the magnifying glass? I would assume that a lot of users would click on it because they would expect a place search functionality...
What about just `Z` or `z`?

Is the change in the styling of the compass (colors) intentional? I prefer how it was before.

orangemug:
> maybe use something different than the magnifying glass? I would assume that a lot of users would click on it because they would expect a place search functionality...
What about just Z or z?

Good point @pathmapper, I guess we should probably see what iconography/conventions other editors use for this. Or maybe just `zoom:` rather than `zoom level:`?

Screenshot 2020-02-18 at 08 00 27  

> Is the change in the styling of the compass (colors) intentional? I prefer how it was before.

Hmm it was intentional, I was just changing back to how it originally was pre `mapbox-gl-js` breaking the CSS overrides. However I think you're right the compass in the before image looks a little clearer. I might make the top arrow of the compass a little lighter somewhere between the before and after screenshots. It that ok?

pathmapper:
> Or maybe just zoom: rather than zoom level:?

Would be fine with me.

I assume your point is to shorten the label to make more of the map visible, which is a good idea.
In fact the zoom level doesn't need to be shown on the top right corner of the map, because it's also shown in the URL hash:

grafik

So I think it's also OK to remove it entirely from the map.

Instead we could also add the `debugToolbox` from the `OL`  to the `GL JS` renderer, which would be also good for consistency between the two renderers.

grafik

So the default would be to grasp the zoom level from the URL hash and if a user prefers the hash information on the map, the `debugToolbox` would fulfill this use case.

If we remove the zoom level from the top right corner, we should add the zoom level to the OL `debugToolbox` because of https://github.com/maputnik/editor/issues/255.

> I might make the top arrow of the compass a little lighter somewhere between the before and after screenshots. It that ok?

Sure, my point is that the north part of the arrow is currently hard to spot.

orangemug:
I think it's worth having a UI element because not everybody will know to look in the URL bar, or which of those numbers represents the zoom. I think it's probably better to have it always displayed because it's one of the key bits of information for map styling, whereas center/rotation don't really come into play when styling. My preference is `zoom: 2.05`, are you happy with that @pathmapper ?

Also worth noting that the URL might not always be present. I have a local branch which makes Maputnik a desktop PWA https://developers.google.com/web/progressive-web-apps/desktop to it appears somewhat like a normal desktop app when 'installed'

> Sure, my point is that the north part of the arrow is currently hard to spot.

Cool, yeah that's what I thought, I'll make some changes and push a new branch for review.

pathmapper:
> Are you happy with that @pathmapper ?

Absolutely 

> I have a local branch which makes Maputnik a desktop PWA

Sounds interesting, and yeah, no address bar :smile:

orangemug:
Updated with discussed changes.

Screenshot 2020-02-18 at 19 00 49

Demo <https://2255-84182601-gh.circle-artifacts.com/0/artifacts/build/index.html#0.25/0/0>

pathmapper:
With the demo the logo in the survey modal is missing:

grafik

Otherwise it looks good.

orangemug:
@pathmapper shall I remove that modal instead? At the moment it opens only for new users and it isn't very relevant for them. The idea when we first introduced it I believe was that old users would see it when they opened the app (once it was released).

Shall we just leave the button in the toolbar instead and remove the modal?

pathmapper:
> Shall we just leave the button in the toolbar instead and remove the modal?

Good idea @orangemug  :+1:

orangemug:
I've temporarily disabled the survey UI, after the next release it might be worth reenabling to popup after the 4-5th time using the app. As a result of possibly re-enabling it I've fixed the image issue highlighted also.

orangemug:
Merging!

PR #618, opened 2020-02-17
Added improved loader with maputnik logo

orangemug:
Uses loader animation from <https://github.com/maputnik/design/blob/master/logos/logo-loading.svg>

Note: the palette dots animate, see demo below
Screenshot 2020-02-17 at 18 58 44

Demo <https://2232-84182601-gh.circle-artifacts.com/0/artifacts/build/index.html#0.36/0/0>

Note: this is only for first load, I see the loader quite a bit during dev and wanted to do this for a while  😁


pathmapper:
Nice @orangemug - out of curiosity: Did you create the SVG by hand or did you use an editor?

orangemug:
Thanks @pathmapper so its a bit of both. How the [maputnik/designs][maputnik-designs] repo works is you have a 'base' SVG that was created in inkscape [src/logos/_includes/base.svg](https://github.com/maputnik/design/blob/master/src/logos/_includes/base.svg) off of the raster that @lukasmartinelli originally created [src/reference/logo.png](https://github.com/maputnik/design/blob/master/src/reference/logo.png). Then the other files reference the parts in that file via

```svg
<!-- you can change the style in the SVG <use/> -->
<use
  xlink:href="_includes/base.svg#map__main"
  fill="#4eba6f"
  stroke="#000000"
/>
``` 

If you start a webserver in the root of the [maputnik/designs][maputnik-designs] repo (you can use `python -m SimpleHTTPServer 8000`) all the logos in <https://github.com/maputnik/design/tree/master/src/logos> will render correctly when opened in the browser. Because external file references when using `<use xlink:href="..." />` don't work everywhere because of browser security (an example is in an `` tag) the build script [`./scripts/build`](https://github.com/maputnik/design/blob/master/scripts/build) resolves/inlines the `xlink:href` in the `<use />` tag.

Note: Yes I should really document that webserver step in the [maputnik/designs][maputnik-designs] `README.md`, on my list.

I think it works ok, and although a little confusing to start with I think it's a really nice setup reusing assets out of a 'base' file. All the animation is done with CSS from within the derived SVG files (for example <https://github.com/maputnik/design/tree/master/src/logos>). I don't think I would setup anything that was likely to change that much this way as it's probably overboard, but for something like a logo where the base design never changes I'm pretty happy with how it works. Although I'd love if an open source SVG editor supported this sort of setup directly.  

It's also worth noting that setting this all up took a while, but I really just wanted to understand SVG a little more so it was as much a learning experience as anything else 😁

[maputnik-designs]: https://github.com/maputnik/design

orangemug:
Merging!

pathmapper:
Thanks a lot @orangemug for the explanation!

Issue #353, opened 2018-07-27
Allow invalid styles and show errors


orangemug:
Editing a Maputnik style can be a pain right now. In many places entering an invalid value will cause the input to get in a funny state or just refuse to let you complete the action. For example changing a color input with the value `rgba(255, 0, 0, 1)` to `rgb(255, 0, 0)` and you'll see the issue.

Also editing a style via the JSON editor is next to impossible at the moment for similar reasons. 

I think we should allow errors to be present in the style, and indicate what layers have errors.

Here's a prototype <https://696-84182601-gh.circle-artifacts.com/0/artifacts/build/index.html#0.26/0/0> (working branch https://github.com/maputnik/editor/compare/master...orangemug:feature/allow-invalid-values)

Below is an example where I've modified a layer `type` to be invalid, `backgroun` instead of `background`

<img width="593" alt="screen shot 2018-07-27 at 17 02 13" src="https://user-images.githubusercontent.com/235915/43332100-4373f27a-91bf-11e8-8d5d-bc12a3736731.png">

Below is an example where I am part way through modifying a color from `hsl` to `hsla` and am still missing the forth argument.

<img width="564" alt="screen shot 2018-07-27 at 17 02 39" src="https://user-images.githubusercontent.com/235915/43332113-4a46f746-91bf-11e8-8656-6b22dad3fff3.png">
 
Any thoughts or feedback welcome and encouraged 😄

gregorywolanski:
This issue appears in the answers to question “What annoys you about Maputnik?” in the survey (original spelling):

- sometimes ctl+v doesnt work if theres is mismatch,it should give error by underline or something

- Black Screen with often no error messge once you make a mistake editing the local style file.

- Bugs to refresh style (I need to play with JSON to fix issues)

pathmapper:
There is already something similar to indicate parse errors:

![auswahl_426](https://user-images.githubusercontent.com/20856381/46808841-874e8700-cd6d-11e8-9f34-274cd78dea11.png)

sk1p:
> Any thoughts or feedback welcome and encouraged

I like it! Maybe in addition to the layer-wide error, also highlight the field that is currently invalid.

pathmapper:
There two regressions in `master` since `v1.5.0` which are likely related: https://github.com/maputnik/editor/issues/484, https://github.com/maputnik/editor/issues/488

orangemug:
Closed by PR #620

Issue #324, opened 2018-06-09
Style errors


orangemug:
Style errors are currently displayed below the map pane and not linked to the field being edited / causing the error.

Being visually far away users could just not notice the error, this has happened to me a few times. I'm also assuming this would be more difficult for users with 'peripheral field loss' (see <https://www.w3.org/TR/low-vision-needs/#h-visfie>).

Tasks

 1. Research how other applications approach this and also what [WCAG](https://www.w3.org/WAI/standards-guidelines/wcag/) advises
 2. Suggest solutions to the issue

orangemug:
Fixed by #620 

Example in screenshot
<img width="368" alt="Screenshot 2020-03-28 at 08 46 13" src="https://user-images.githubusercontent.com/235915/77819169-9bf11d00-70d0-11ea-960a-53346e8ed179.png">

PR #606, opened 2020-01-19
Added style formatting into the api store

orangemug:
This is a fix for #591. However at the moment that part is untested as I don't personally have maputnik/desktop setup and working. If someone has maputnik/desktop setup and wants to test the build, feel free. Otherwise I'll get round to it next week, when my brain is a little fresher

PR #620, opened 2020-02-18
Added support for expressions and UI errors

### LINKED ISSUE #223: Add support for expressions
This ticket is here to start discussions around [expressions](https://www.mapbox.com/mapbox-gl-js/style-spec/#expressions). See <https://blog.mapbox.com/announcing-expressions-in-gl-js-a72b55d0a6af> for a good introduction to the subject.

Questions we need to answer

 - How are users currently using expressions, is the JSON editor sufficient?
 - If not how would you like to see expressions implemented?

Currently unsupported features, due to lack of expression support

 - `line-gradient`
 - `*-sort-key`

orangemug:
This PR adds support for [mapbox-gl-js expressions](https://docs.mapbox.com/mapbox-gl-js/style-spec/expressions/). Via the new `Fx` button

Screenshot 2020-02-18 at 08 15 15

You can also revert/trash an expression via the 

Screenshot 2020-02-18 at 08 16 00

Also you can convert existing zoom/data functions to expressions

Screenshot 2020-02-18 at 08 27 55

Adding support for expressions without inline errors didn't make much sense as making errors while editing a blob of JSON is quite common. So this PR also adds

Inline errors

Screenshot 2020-02-18 at 08 16 15

Layer errors displayed in layer list

Screenshot 2020-02-18 at 08 16 25

Error bar + 'switch to layer' link

Screenshot 2020-02-18 at 08 16 47

This PR also adds functionality to render what it can of a broken style. So for example if you have 30 layers and one of those layers has an invalid property value, for example `background-color`, it'll remove that single property from the style before sending it to mapbox-gl/open layers renderers. This allows you to fix a broken style in stages.

I **think** this is almost complete, I'm aware of some issues with undo/redo (ctrl+z) when changing to an expression from a normal value. But basically it needs testing to make sure it 'feels right'. So please have a try and tell me what you think. I also need to tidy the code a little and add a few comments. 

The original testing, conversation was happening in #223 so you may also want to look there for any more context. 

Demo: <https://2252-84182601-gh.circle-artifacts.com/0/artifacts/build/index.html> 

Fixes #223 #353 

Some work towards #324


pathmapper:
@orangemug overall this is great. 

Some legacy filters are treated as expressions, e.g.

`["==", "class", "residential"]`

from `OSM Liberty` style, layer `landuse_residential`.

Also in this case there is no inline error:

Peek 2020-02-22 10-48

orangemug:
@pathmapper thanks, so this was something I was struggling with and forgot about... oops. So to explain a little about why it's awkward. The style spec has one property but effectively 2 types of data are allowed 'filter' (the old style) and 'expression' (the new style). Where it becomes awkward is it's very easy to accidentally write a valid 'filter' while you're editing an 'expression'. Originally during development I had it auto-switch to the 'filter' UI when it was a valid filter, which was very annoying. The following comment gave me an idea

> Also in this case there is no inline error:

So if we encounter a valid 'filter' when we've switched to 'expression' mode rather than displaying an 'error', because it is technically a valid property value (just not an expression) we instead display an info box, telling the user about this.

Screenshot 2020-02-22 at 11 54 28

That then gives the user a few possible options

 - If they reached this point while editing an 'expression' then they can just ignore the 'info box'
 - If they would like to edit the value in a filter UI they have the option to switch
 - Else they can just continue editing the JSON

It's worth noting that a filter will always be displayed with the filter UI on first load of the layer, but I don't see that as a major issue because users wanting to edit the 'filter' as JSON can use the 'JSON editor' at the bottom of the layers panel

Let me know what you think, demo <https://2298-84182601-gh.circle-artifacts.com/0/artifacts/build/index.html#0.36/0/0>

pathmapper:
@orangemug good idea! 

Now it looks like the inline errors and the layer errors displayed in layer list are not working anymore.

image

pathmapper:
@orangemug we are having currently two different types of errors - JSON parse errors and style spec validation errors, right?

I think the parse errors are blocking and further style spec validation and update of the JSON editor and are also not displayed in the layers list.

Would it be possible to handle the parse errors the same was as the style spec validation errors?

pathmapper:
Or is removing the codemirror linting an option because the style spec validation checks also for syntax errors?

orangemug:
@pathmapper did you mean something like the following?

Screenshot 2020-02-23 at 15 45 43

Another oddity is validation between 'filter' and 'expression'. It looks like in the following instance _mapbox-gl_ is validating it as an 'filter' by the top style validator function, it's a little confusing to the user (and to me).  

Screenshot 2020-02-23 at 15 56 49

I'm thinking it's probably better that I make the JSON editor use the new expression syntax only. The added advantage is it looks like the expression validation always produces 1 error rather than multiple (not 100% about this), whereas the filter errors can produce multiple, as seen in the above screenshot.

pathmapper:
@orangemug 

> did you mean something like the following?

Yep, :+1: 

> It looks like in the following instance mapbox-gl is validating it as an 'filter' by the top style validator function

I think in this instance the validation is the same for the deprecated 'filters' and 'expression' and _filter_ has to meanings here:
1. all together ["all", "foo", "bar"]
2. and also two filters which are combined into one filter ["all", `filter1`, `filter2`]

Expressions are represented as JSON arrays, so it expects an array for `filter1` and `filter2`.

If you run `gl-style-validate` on the command line on a style with this instance, the validation will look like this:

```
$ gl-style-validate osm_liberty_fb.json 
osm_liberty_fb.json:65: layers[4].filter[1]: array expected, string found
osm_liberty_fb.json:65: layers[4].filter[2]: array expected, string found
```

> I'm thinking it's probably better that I make the JSON editor use the new expression syntax only. The added advantage is it looks like the expression validation always produces 1 error rather than multiple (not 100% about this)

If this is true, I'm all for it. But I doubt that it is working like this.

pathmapper:
@orangemug I've did some more tests and I think you are right with 

> in the following instance mapbox-gl is validating it as an 'filter'

How do you want to achieve: 

> make the JSON editor use the new expression syntax only 

Create a copy of the style spec without 'filter' and use this custom spec for validation?

orangemug:
Thanks @pathmapper, I think there is a method exposed from `@mapbox/mapbox-gl-style-spec` for validating expressions. I'll take a look later on tonight/tomorrow.

In the meantime I've added the JSON parse error to the UI, see <https://2336-84182601-gh.circle-artifacts.com/0/artifacts/build/index.html#0.25/0/0>

orangemug:
Apologies for the delay it's been a busy week. I've made some progress, using the errors from `createExpression` exported from the `mapbox-gl-style-spec` I've added inline JSON errors

Screenshot 2020-02-29 at 10 31 27

This is all local at the moment as it needs some tidying up and a couple of bugs fixed, so the code won't be pushed to GitHub until next week. After I push this code I'll also look to add the same 'inline errors' to the JSON editor at the bottom of the layer pane, as it should be pretty much the same thing.

pathmapper:
Thanks for the update @orangemug 

> to add the same 'inline errors' to the JSON editor at the bottom of the layer pane

Sounds like this PR would then fully adress https://github.com/maputnik/editor/issues/324 :+1:

orangemug:
It still requires some testing from me, however the latest version of expression support is pushed with the previously mentioned errors inline in the JSON editor

Screenshot 2020-03-08 at 18 37 32

There is almost definitely some bugs and a little bit of tidying is still required. However give it a general try list me know how it feels.

Try it at <https://2345-84182601-gh.circle-artifacts.com/0/artifacts/build/index.html>

orangemug:
@pathmapper I've had a bit of a tidy and another test of this feature and it looks stable now to me, would you mind giving it a test?

pathmapper:
> would you mind giving it a test?

Thanks @orangemug, I'll try to find some time this week to have a look.

pathmapper:
@orangemug did some testing and so far it works great.

orangemug:
Thanks @pathmapper. Would you like more time to test, or shall we merge?

pathmapper:
Let's merge @orangemug

orangemug:
Thanks @pathmapper, merging 🎉

This is a big update, thanks for all the help/feedback/support/testing along the way. I know it took a while but I think we have a good solution here. I'll roll this into a beta release sometime over the next week along with #606 then we can get a new release out there in the world.

pathmapper:
> I think we have a good solution here

Yep, think so, too. Great work on this @orangemug :tada:

PR #625, opened 2020-03-28
Fix UI when loading invalid style with duplicate layer ids

### LINKED ISSUE #401: Add /edit layer - Error is missing if multiple layer IDs are an empty string
Currently it is possible to add a Layer without ID which is [required and must be unique](https://www.mapbox.com/mapbox-gl-js/style-spec#layer-id).

orangemug:
This PR fixes the UI when the style is in an invalid state where there are 2 layers with the same `id`. Previously you could only select/edit one of the layers with the duplicate ids. This changes the editing to work off of the layer `index` rather than layer `id`

This also adds the missing error from validate where two layers with the id `""` (empty string) are in the style, showing the appropriate error.

Demo <https://2379-84182601-gh.circle-artifacts.com/0/artifacts/build/index.html>

References #401


pathmapper:
Thanks @orangemug, if I add two background layers with an empty string as name, there is no in inline error as I would expect like the error for the color property in the screenshot of https://github.com/maputnik/editor/issues/324#issuecomment-605415152:

grafik

orangemug:
Hey @pathmapper, so there are some reasons for this which I should have explained in the original ticket (sorry). So if you create 2 layers with the same id for example `test` you only get an error in the error pane below the map, this is because I don't get a line number with the validation error. So this ticket brings the `[empty_string]` use case inline with that.

The error is actually 

```
layers[1]: duplicate layer id "test", previously used at line undefined
```

The `undefined` comes straight from the `@mapbox/mapbox-gl-style-spec` so we probably have a few issues to raise in the upstream repo.

Do you think it makes sense to bring `[empty_string]` inline with the current error? I agree it generally needs fixing but probably better to be fixed upstream.

pathmapper:
> I don't get a line number with the validation error

Hmm, if I use `gl-style-validate` from `@mapbox/mapbox-gl-style-spec` on this [style](https://github.com/maputnik/editor/files/4399033/osm_liberty.zip) I get 

`osm_liberty.json:1684: layers[103]: duplicate layer id "test", previously used at line 1683`

So there is a line number available.

orangemug:
@pathmapper, hmm..... I will have to double check I'm using the library correctly then 😀

Leave it with me, I'll give it another look.

orangemug:
Hey @pathmapper fixed in latest commit, see <https://2399-84182601-gh.circle-artifacts.com/0/artifacts/build/index.html>

Yeah I was getting myself all confused, for clarification `gl-validate-style` reads in a string hence it knows the line number. In our code we generate an AST (abstract syntax tree) from the JSON, and reference the error in that deeply nested object. I just forgot what I'd previously coded... 🙄

pathmapper:
Works great @orangemug, nice solution with "switch to layer" :tada: 

grafik

orangemug:
🎉merging!

Issue #320, opened 2018-06-09
Keyboard accessible layer menu


orangemug:
The layers menu is not currently keyboard accessible. I believe the correct approach would be to use an aria 'tree view' component, see:

 - https://www.w3.org/TR/2017/NOTE-wai-aria-practices-1.1-20171214/examples/treeview/treeview-1/treeview-1b.html
 - http://accessibleculture.org/articles/2013/02/not-so-simple-aria-tree-views-and-screen-readers/

We need a standalone react component we can just `npm install`

Tasks

 1. Look for existing components we can use
 2. If we can't find an existing component, make one of our own

PR #626, opened 2020-03-28
Moved to using react-accessible-accordion for <LayerEditorGroup/>

### LINKED ISSUE #321: Accessible collapsable layer groups
The collapsable 'layer groups' when editing a layer are not currently keyboard accessible.

There is an existing pull request for this <https://github.com/maputnik/editor/pull/299>. However as it says in that ticket <https://github.com/springload/react-accessible-accordion> is probably a better approach.

orangemug:
This does remove the animation, but that seems like a small price for accessibility, we can work out how to put back in animation later.

Screenshot 2020-03-28 at 12 58 04

Demo <https://2383-84182601-gh.circle-artifacts.com/0/artifacts/build/index.html>

Fixes #321


pathmapper:
> that seems like a small price for accessibility, we can work out how to put back in animation later.

@orangemug agree, what about 
https://github.com/maputnik/editor/blob/master/src/components/layers/LayerListGroup.jsx ?

orangemug:
`<LayerListGroup/>` is slightly different it needs to be a 'tree view', see #320. I've got something partially working for that but it's slight more involved.

pathmapper:
Thanks @orangemug, looks good to merge.

orangemug:
Thanks, merging!

Issue #630, opened 2020-04-02
Scroll to layer list item after click on layer in <FeatureLayerPopup />


Jinkwon:
I found very simple UI issue here.

1. Click map feature and see selected layer list popup.
2. And then clicking some layer.
3. Changed selected layer view left side, but I couldn't found layer from list (but it changed focused item)

If scroll to selected layer from list, it makes sense.
Please consider for it.

<img width="987" alt="스크린샷 2020-04-02 오후 7 20 01" src="https://user-images.githubusercontent.com/1798916/78237989-ffd37580-7516-11ea-9955-ecac2cdb9794.png">

pathmapper:
Thanks for the suggestion @Jinkwon, this would be indeed a nice improvement.

orangemug:
Yeah much better, thanks for the suggestion @Jinkwon. It's also (I assume) a pretty big improvement for those with certain learning difficulties. 

Fixed in #634

PR #639, opened 2020-04-06
Only scroll to selected item in <LayerList/> if not already in view.

orangemug:
Fix for 'jumping' `<LayerList/>` introduced in #630.

The issue was that we'd always call `element.scrollIntoView()` in certain circumstances if the selected item was already in scrollable view the layer would change scroll position.


orangemug:
AppVeyor still flakey, CircleCI test passing. Merging!

Issue #641, opened 2020-04-10
New Thunderforest tilesets


gravitystorm:
A few months ago we released upgraded versions of our vector tilesets ([1](https://www.thunderforest.com/blog/vector-outdoors-v2/), [2](https://www.thunderforest.com/blog/vector-transport-v2/)). It would be great to upgrade them in Maputnik too.

The new tilesets are (unsurprisingly!):

* [`thunderforest.outdoors-v2`](https://www.thunderforest.com/docs/thunderforest.outdoors-v2/)
* [`thunderforest.transport-v2`](https://www.thunderforest.com/docs/thunderforest.transport-v2/)

There are lots of improvements in these versions of the tilesets, including significant simplifications for improved tile sizes, and reworked layers for easier cartography. Hopefully they'll be even more useful than our `v1` tilesets! The `v1` tilesets are still available, but there's no reason to use them for new projects.

pathmapper:
That's great, thanks for letting us know @gravitystorm.

Upgrade done in https://github.com/maputnik/editor/pull/642 which will be included in the upcoming `v1.7.0` release of Maputnik.

orangemug:
The new tilesets are looking really nice @gravitystorm, also looking a lot smaller than before, congrats 🎉

It's probably worth linking out to your docs or announcement in the release blog post for maputnik `v0.7.0`. Do you think linking to https://www.thunderforest.com/blog/vector-transport-v2/ and https://www.thunderforest.com/blog/vector-outdoors-v2/ makes the most sense?

orangemug:
Oh also @gravitystorm if you have any open source styles for thunderforest vector sources, we can add them to the 'open modal'

gravitystorm:
> Do you think linking to https://www.thunderforest.com/blog/vector-transport-v2/ and https://www.thunderforest.com/blog/vector-outdoors-v2/ makes the most sense?

Yep, those are probably the best links

> Oh also @gravitystorm if you have any open source styles for thunderforest vector sources, we can add them to the 'open modal'

Not yet, but hopefully soon! I'll open an issue when we've got some available.

> also looking a lot smaller than before, congrats

Thanks! Do you think it's worth revising the warning about the tilesets being "heavy"? I know that for some layers they are about 80% smaller than the v1 tiles, but what do you think about them overall?

orangemug:
[...]

PR #644, opened 2020-04-12
Remove notes from thunderforest sources

orangemug:
Thunderforest tiles are no longer 'heavy' as of v2 🎉(see https://github.com/maputnik/editor/issues/641#issuecomment-612593093)

Removed all references from 'heavy' from the UI.


pathmapper:
Yep, I'm currently playing around with the v2 tiles, huge difference compared to v1 :tada:

PR #645, opened 2020-04-12
Add support for identity functions

### LINKED ISSUE #640: Add support for identity functions
First raised by @pathmapper in https://github.com/maputnik/editor/issues/629#issuecomment-609777086

We don't currently have support for [identity functions](https://docs.mapbox.com/mapbox-gl-js/style-spec/other/#function-type) 

image

So I think we can add it to the data functions UI, we'll add another option to the `<select/>` box labeled `identity` and it'll just have a property value, like the following (but with `identity` instead of `exponential`)

Screenshot 2020-04-06 at 20 07 07

Thoughts?

orangemug:
I've added _identity function_ support, screenshot below

Screenshot 2020-04-12 at 16 29 25

You can also undo from a `["get", "property_name"]` expression back to a _identity function_.

Demo <https://2490-84182601-gh.circle-artifacts.com/0/artifacts/build/index.html>

Fixes #640

### DIFF
--- a/src/components/fields/FunctionSpecField.jsx
+++ b/src/components/fields/FunctionSpecField.jsx
@@ -6,10 +6,19 @@ import DataProperty from './_DataProperty'
 import ZoomProperty from './_ZoomProperty'
 import ExpressionProperty from './_ExpressionProperty'
 import {function as styleFunction} from '@mapbox/mapbox-gl-style-spec';
+import {findDefaultFromSpec} from '../util/spec-helper';
 
 
 function isLiteralExpression (value) {
   return (Array.isArray(value) && value.length === 2 && value[0] === "literal");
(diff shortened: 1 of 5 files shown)


pathmapper:
Works great @orangemug :+1: 

Should we also add `Fx` to the "Convert to expression" button to have consistency with all the other fields?

orangemug:
@pathmapper do you mean something like this?

Screenshot 2020-04-14 at 08 29 15

I also add a 'add row' icon to the 'add stop', thoughts?

pathmapper:
Yes, that is exactly what I had in mind @orangemug

The 'add stop' looks also nice :tada:

orangemug:
Yay merging!

PR #650, opened 2020-04-13
Fixed crash raised in issue #647

### LINKED ISSUE #647: Uploading style crashes the editor while in inspect mode
Steps to reproduce:
1. Use https://maputnik.github.io/releases/v1.7.0-beta2/
2. Open the `OSM Liberty` style
3. Switch to inspect mode
4. Unzip and upload the following style [test_housenumber_with_font.zip](https://github.com/maputnik/editor/files/4468222/test_housenumber_with_font.zip)
5. The editor crashes with the following error:
grafik

orangemug:
Demo <https://2498-84182601-gh.circle-artifacts.com/0/artifacts/build/index.html>

Fixes #647

### DIFF
--- a/src/components/map/MapboxGlMap.jsx
+++ b/src/components/map/MapboxGlMap.jsx
@@ -110,11 +110,11 @@ export default class MapboxGlMap extends React.Component {
       this.state.inspect._popupBlocked = false;
       this.state.inspect.toggleInspector()
     }
-    if(this.props.inspectModeEnabled) {
-      this.state.inspect.render()
-    }
-
     if (map) {
+      if (this.props.inspectModeEnabled) {
+        this.state.inspect.render();
+      }
+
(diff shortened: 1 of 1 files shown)


pathmapper:
@orangemug the editor is still crashing, but now with a different error:

grafik

orangemug:
@pathmapper it's a bit of a hack, looks to be an issue with mapbox-gl-inspect, I need to look into it further but the hack add in the previous commit seems to do the job. I think the map is in an odd state at that point, just catching the error doesn't appear to cause any side effects.

Demo <https://2518-84182601-gh.circle-artifacts.com/0/artifacts/build/index.html>

pathmapper:
@orangemug I can confirm that the crash is not reproducible anymore. Also I don't see any side effects, so this looks good to merge.

PR #654, opened 2020-04-13
Fixed default values for <FontInput/>

### LINKED ISSUE #649: Default value for text-font property is not displayed correctly
Steps to reproduce:
1. Use https://maputnik.github.io/releases/v1.7.0-beta2
2. Unzip and open the following style: [test_housenumber.zip](https://github.com/maputnik/editor/files/4468270/test_housenumber.zip)
3. There is only `Arial Unicode MS Regular` displayed in the text font field.
The default should be _["Open Sans Regular","Arial Unicode MS Regular"]_ according to the [spec](https://docs.mapbox.com/mapbox-gl-js/style-spec/layers/#layout-symbol-text-font).

grafik.

If you switch to expression syntax, the font array is as expected:
grafik

orangemug:
Fixes default values for `<FontInput/>`, it was previously was only picking up the first value of the array of values.

Demo <https://2502-84182601-gh.circle-artifacts.com/0/artifacts/build/index.html>

Fixes #649

### DIFF
--- a/src/components/inputs/FontInput.jsx
+++ b/src/components/inputs/FontInput.jsx
@@ -4,7 +4,7 @@ import AutocompleteInput from './AutocompleteInput'
 
 class FontInput extends React.Component {
   static propTypes = {
-    value: PropTypes.array.isRequired,
+    value: PropTypes.array,
     default: PropTypes.array,
     fonts: PropTypes.array,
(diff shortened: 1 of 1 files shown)


pathmapper:
Looks good :+1:

PR #663, opened 2020-04-21
Fix for desktop build

nspringer-trimble:


orangemug:
Thanks @nspringer-trimble merging!

Also hello and welcome to Maputnik, if you have a moment, please tell us what you use Maputnik for over in #164

PR #666, opened 2020-04-25
Fix to make layer list header visible at all times

### LINKED ISSUE #646: Layer list: Head should be visible all the time
Currently, if there are many layers listed, it happens that the head scrolls away and there is a need to scroll back to the top if you want to use the "Expand/Collapse" and  "Add Layer" buttons:

image

buttons_scroll_away

orangemug:
I've made the layer list header always visible and also improved scrollbar styling/positioning for toolbar.

Demo <https://2547-84182601-gh.circle-artifacts.com/0/artifacts/build/index.html>

Before - note that I've made the window small so the top toolbar overflows also
Screenshot before changes

After
Screenshot after changes

Fixes #646

### DIFF
--- a/src/components/Toolbar.jsx
+++ b/src/components/Toolbar.jsx
@@ -210,7 +210,11 @@ export default class Toolbar extends React.Component {
           <ToolbarSelect wdKey="nav:inspect">
             <MdFindInPage />
             <IconText>View </IconText>
-            <select onChange={(e) => this.handleSelection(e.target.value)} value={currentView.id}>
+            <select
+              className="maputnik-select"
+              onChange={(e) => this.handleSelection(e.target.value)}
+              value={currentView.id}
+            >
               {views.map((item) => {
                 return (
                   <option key={item.id} value={item.id} disabled={item.disabled}>
--- a/src/styles/_input.scss
+++ b/src/styles/_input.scss
@@ -97,6 +97,13 @@
 .maputnik-select {
   @extend .maputnik-input;
 
+  -moz-appearance: none;
+  -webkit-appearance: none;
+  background: $color-gray url("#{$icon-down-arrow}") right center no-repeat;
+  color: $color-white;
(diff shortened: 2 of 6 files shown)


pathmapper:
With this PR and Firefox/Linux the dropdown buttons became very tiny:

grafik

This is how it looks with `v1.7.0`:
grafik

pathmapper:
After some digging I think there are some general styling issues regarding the select box on FF, not only on Linux, also on MacOS and likely also on other OS.

https://support.mozilla.org/de/questions/1266036
https://bugzilla.mozilla.org/show_bug.cgi?id=1571386


E.g. when you remove the `border-width` property from the select, you'll get a nice looking dropdown arrow with FF on Linux but other styling issues:

grafik

From 

Accoring to the following information, the way to go is to remove the https://developer.mozilla.org/en-US/docs/Learn/Forms/Advanced_form_styling#Selects_and_datalists

pathmapper:
@orangemug maybe merge this PR and do the `select` styling in a separate one because it will affect also the `selects` of `<LayerEditorGroup/>` and the `Sources` and `Settings` modals?

orangemug:
@pathmapper what do you think?

Demo <https://2562-84182601-gh.circle-artifacts.com/0/artifacts/build/index.html>

Chrome (MacOS)
Screenshot 2020-04-26 at 09 24 18

Firefox (MacOS)
Screenshot 2020-04-26 at 09 24 39

pathmapper:
Cool @orangemug, looks the same with FF/Chrome on Linux :tada: 

If you could move the view/select part a little bit up to the same bottom line as the other toolbar items, it would be perfect.

grafik

orangemug:
Fixed alignment <https://2571-84182601-gh.circle-artifacts.com/0/artifacts/build/index.html>

pathmapper:
Perfect, thanks!

PR #683, opened 2020-05-18
Fix some accessibility issues

### LINKED ISSUE #320: Keyboard accessible layer menu

orangemug:
This pull request improves the editor for the following users

 - keyboard only users
 - those with visual impairments, who use a screen reader for some, although not all of their editing

Although many of them are generally useful to people outside those groups also.

### Features/fixes
Added aria landmarks to describe the main areas of the editor. With MacOS 'voice over' enabled pressing `CTRL + ALT + U` you'll see

Voiceover landmark menu with the Maputnik editor

Added missing title attributes to all icon only buttons. Both mouse users when hovering the button and screen readers will get descriptive text.

Icon only button showing title attribute on hover

Inaccessible `<Multibutton/>` component has been changed internally to be a group of radio buttons, whereas before it was `<button>`s. This improves screen reader accessibility, because it is now announced as a group. Also keyboard users can access it using the arrow keys, as they would with normal radio buttons (up/down or left/right)

<img width="385" alt="Screenshot of 2 <Multibutton/> components" src="https://user-images.githubusercontent.com/235915/82249628-0b052800-9942-11ea-9918-4833c0a319d7.png">

Replaced 1 'skip navigation link' with 3 more obvious 'skip links'

 - Layers list
 - Layer editor
 - Map view

Tab based keyboard interaction (versus using something like voice over) was really slow previously because you had to tab through a lot of links/buttons to get to what you wanted to interact with. With the added 'skip links' a user can navigate back to the URL bar (usually `CTRL + L` / `CMD` + L`) and restart navigation from the quick links.

Visible display of layers list skip link

Added the existing map specific shortcuts to the shortcuts menu (`?` menu). The shortcuts menu also compliments the previous keyboard improvements.

Screenshot 2020-05-18 at 18 10 25

I've hidden 'layer list' actions from tab order, preferring to use the layer editor options. Having 3 buttons per menu item, increased the amount of tab navigation you needed and the 'layer editor' menu is a lot more efficient.

Outlined image showing layer list/layer editor options

Also the 'tree view' approach for the 'layer list' as mentioned in #320 was a no-go, because as mentioned at https://dequeuniversity.com/library/aria/tree-view it's currently not supported by screen readers

> WARNING: Screen Reader support for tree components is spotty at best. 

I instead went with a 'menu button' aria markup approach instead, similar to the 'layer editor' more menu. 

### Summary
Hopefully the above improves the editing experience for everyone. I know personally using a mix of keyboard/mouse is my preferred method of using most applications.

I will be continuing to work on improving accessibility, I'll expand on further work to be done in other tickets.

If anyone reading this ticket has any accessibility requirements that are not met by the current editor, please reach out and let us know. We can't promise to fix everything, but we'll always try to help those who struggle.

Fixes #320

### DIFF
--- a/src/components/Button.jsx
+++ b/src/components/Button.jsx
@@ -13,6 +13,7 @@ class Button extends React.Component {
     disabled: PropTypes.bool,
     type: PropTypes.string,
     id: PropTypes.string,
+    title: PropTypes.string,
   }
 
   render() {
--- a/src/components/Button.jsx
+++ b/src/components/Button.jsx
@@ -12,10 +12,13 @@ class Button extends React.Component {
     children: PropTypes.node,
     disabled: PropTypes.bool,
     type: PropTypes.string,
+    id: PropTypes.string,
   }
 
   render() {
     return <button
+      id={this.props.id}
+      title={this.props.title}
       type={this.props.type}
       onClick={this.props.onClick}
       disabled={this.props.disabled}
--- a/src/components/Toolbar.jsx
+++ b/src/components/Toolbar.jsx
@@ -131,6 +131,16 @@ export default class Toolbar extends React.Component {
     this.props.onSetMapState(val);
   }
 
+  onSkip = (target) => {
+    if (target === "map") {
+      document.querySelector(".mapboxgl-canvas").focus();
+    }
+    else {
+      const el = document.querySelector("#skip-target-"+target);
+      el.focus();
+    }
+  }
+
   render() {
     const views = [
       {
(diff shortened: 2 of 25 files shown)


pathmapper:
Demo link for anyone who wants to try this out:
https://2606-84182601-gh.circle-artifacts.com/0/artifacts/build/index.html

orangemug:
I'm going to merge this now, if anybody has any comments please feel free to comment post merge or open a new ticket.

Issue #322, opened 2018-06-09
Accessible JSON editor


orangemug:
The JSON editor isn't currently accessible to screen readers. Switch to an accessible editor, probably [monaco-editor](https://github.com/Microsoft/monaco-editor) see <https://github.com/Microsoft/monaco-editor/wiki/Accessibility-Guide-for-Integrators> for details.

I gave this a try and it seems the `monaco-editor-webpack-plugin` only supports webpack 4. So probably best complete/merge #316 first

orangemug:
We can also switch to [codemirror v6](https://codemirror.net/6/)

> The project is at a point where the core code is in good shape, but the documentation is incomplete and several essential extensions have not been implemented yet.
From the codemirror 6 website

I think it's worth giving codemirror a try, because

 a. codemirror is really nice to use
 b. its been around for a long time now, and I feel like codemirror has a greater likelihood of being around 3 years from now and still maintained

orangemug:
This requires a bit more investigation from me, I need to understand the trade offs as there seems to be bugs with both codemirror-v6 and monaco-editor. I need to understand a little more first though about screen readers, so pushing this back to v1.9.0

PR #686, opened 2020-05-21
Added tip to JSON editor about how to unfocus

orangemug:
The JSON editor now displays a helper message for a couple of seconds if you keyboard tab to it.

> Press `ESC` to lose focus

This is _only_ present when focusing via keyboard (tab) and not mouse 

Screenshot 2020-05-21 at 09 13 16

Related to #322

### DIFF
--- a/src/components/layers/JSONEditor.jsx
+++ b/src/components/layers/JSONEditor.jsx
@@ -159,7 +159,7 @@ class JSONEditor extends React.Component {
       style.maxHeight = this.props.maxHeight;
     }
 
-    return <div className="JSONEditor" onPointerDown={this.onPointerDown}>
+    return <div className="JSONEditor" onPointerDown={this.onPointerDown} aria-hidden="true">
       <div class={classnames("JSONEditor__message", {"JSONEditor__message--on": showMessage})}>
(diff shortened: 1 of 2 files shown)


src/components/layers/JSONEditor.jsx
```diff
@@ -145,16 +154,22 @@ class JSONEditor extends React.Component {
...
       style.maxHeight = this.props.maxHeight;
     }
 
-    return <div
-      className={classnames("codemirror-container", this.props.className)}
-      ref={(el) => this._el = el}
-      style={style}
-    />
+    return <div className="JSONEditor" onPointerDown={this.onPointerDown} aria-hidden="true">
+      <div class={classnames("JSONEditor__message", {"JSONEditor__message--on": showMessage})}>
```
pathmapper, line 164: Looks like the failing tests could be fixed with:
```suggestion
      <div className={classnames("JSONEditor__message", {"JSONEditor__message--on": showMessage})}>
```

orangemug: I just realised that was a 'Suggested change' and I could have fixed that from the UI rather than my editor. Thats cool, didn't know you could do that 🤦‍♀️

Issue #319, opened 2018-06-09
Input field accessibility


orangemug:
Input elements are not currently associated with their labels. Therefore they are not accessible to a screen reader and break a general UX pattern of not being able to click on a label to focus the input.

There are two ways to achieve this in the DOM

```html
<label for="some-unique-id"></label>
<input id="some-unique-id" />
```

or

```html
<label>
  <input />
</label>
```

orangemug:
So an update with some research, so looking at https://webaim.org/techniques/forms/controls we need to markup our field differently depending on what's in them, examples from https://webaim.org/techniques/forms/controls follow

`<FieldMultiInput/>` should look something like

```html
<fieldset>
  <legend>Translate anchor</legend>
  <label>
    <input type="radio" value="viewport">
    Viewport
  </label>
  <label>
    <input type="radio" value="map">
    Map
  </label>
</fieldset>
```

Whereas `<FieldString/>` should look something like

```html
<label for="id">ID</label>
<input id="id" />
```

The issue is at the moment we have a single container element called `<Block/>` which contains the labels/actions and the children of that component is the <Field* />` components. So for example

```jsx
<Block label={"ID"} fieldSpec={latest.layer.id}
  data-wd-key={this.props.wdKey}
  error={this.props.error}
>
  <FieldString
    value={this.props.value}
    onChange={this.props.onChange}
  />
</Block>
```

I'm going to get rid of the generic `<Block/>` element instead adding in the `fieldSpec`/`error` into the `<Field* />` so for example the above would be rewritten as

```jsx
<FieldString
  label={"ID"}
  fieldSpec={latest.layer.id}
  value={this.props.value}
  onChange={this.props.onChange}
/>
```

Which kind of makes sense because in some `<Field*/>` components we already pass in fieldSpec anyway, for example `<FieldDynamicArray/>`

orangemug:
@pathmapper I've got a little further with this (urgh it's a fair bit of work). Basically the difficulties are that most of the input fields are inaccessible. I think I've fixed most of the issues now and will post a PR hopefully Monday.

There is one aspect that was a little more difficult, that was 'functions', so the issue is that the UI isn't really that self explanatory for a user with perfect sight because the UI is lacking a lot of labels. So I though I'd have a go at redesigning it. Here's what I've come up with

**Before**
[...]

PR #699, opened 2020-06-01
Tidy of components + added storybook.js

orangemug:
Improvements

 - Moved all components into a single directory like next.js does. The components we getting really messy and the sub-directories weren't helping, this is a push to make it easier. There is still some work to do, but this is a good start
 - Made component names consistent with each other, they all prefix rather than suffix `<FieldNumber/>`, `<FieldArray/>` for example
 - Made component names consistent with their export class names. Previously the `class IconLayer {}` wasn't always consistent with the filename of the component
 - Added storybook for a few components with the aim to extend this across all the fields, and maybe further (see screenshot)
 - Add storybook as a CI artifact

Storybook allows for easier improvement/testing of existing/new components in a standalone way. Here's a screenshot of the current components. 

Screenshot 2020-06-01 at 16 24 56
(Note: Missing fonts is a known issue, will fix that in a later PR)

---

You can see storybook artifact in https://github.com/maputnik/editor/actions/runs/121594306


orangemug:
I've added a bunch more components to the storybook and also the [a11y addon](https://github.com/storybookjs/storybook/tree/master/addons/a11y)

Updated screenshot of storybook

_Note:_ Some of the accessibility errors aren't actually errors, but things that don't make sense in the current context of the component. Currently I'm just ignoring those at the moment, but later on we should find out a way to disable them.

orangemug:
Merging to make it easier to start work on #319, any feedback please leave in this ticket and I'll look at it retrospectively

PR #706, opened 2020-06-10
Improved label accessibility

orangemug:
Most of the inputs are a lot more accessible now. Where there are lists in the UI for example (dynamic) arrays of values these are now marked up as lists, so `<ul/>` & `<li/>`.

Labels are now associated with their inputs which makes using keyboard tab-navigation a lot nicer, before you'd get a keyboard focus for the `<label/>` and `<input/>`, now we just get a tab-navigation for the `<input/>`, however the input is still described by the label for screenreaders.  

screenshot of editor with inputs associated with labels

Where we have groups of elements like data/zoom functions or radio buttons they are marked up as fieldsets. We also have the updated UI for both data and zoom functions.

zoom function screenshot
data function screenshot

There are now [storybook](https://storybook.js.org/) stories for most of the components in the UI, we now have over 40 stories.

During testing it was becoming a pain point continually having to navigate through the UI on page refresh between changes, although the standalone storybook helped somewhat, when I needed to dev in the actual editor it was always a pain. I've added a search param based router to the editor to help with this. It's of the format

```
http://localhost:8080/editor/?layer=[hash_of_style]~[layer_index]&view=[map_view]&modal=[modals_that_are_open]
```

For example if you select layer index 3, set map view to 'inspect' and open the 'export' modal. The URL would look like the following, where `2995677221` is a hash of the style json.

```
http://localhost:8080/editor/?layer=2995677221~3&view=inspect&modal=export
```

Which means if you refresh the page the UI will return to the state it was previously, which is really speeding up development. It could also be useful for creating some docs in the future as you can link to a UI state. Although that not the priority here, so we might end up disabling it pre-merge to address it (test it) for the next release.

I've also been through the UI with the [axe chrome accessibility extension](https://chrome.google.com/webstore/detail/axe-web-accessibility-tes/lhdoppojpmngadmnindnejefpokejbdd) and fixed most of the issues reported the remaining issues are 

 - 'ARIA hidden element must not contain focusable elements' - which will be resolved in #322 
 - 'Elements must have sufficient color contrast' - which we should look to resolve in the next release
 - 'id attribute value must be unique' - this is just an issue with the Maputnik SVG logo, I need to investigate wether it's worth fixing.

Screenshot of accessibility results

I now also mention 'color accessibility' in the view selector

Color accessibility described in view selector

Still left todo

 - Add back info button and SDK docs to fields
 - Fix some styling issues
   - Data sources modal
   - Fix some wrapping inputs in various places, most obvious is the settings modal 
 - Give it all another good test


orangemug:
🎉 Ready for review

Also @pathmapper do you think we should keep the new URL router? You'll notice the querystring changes as you open modals, navigate through layers and edit the style. For example if you open https://mzh48.sse.codesandbox.io/?layer=1078604094~0&modal=sources it'll open the editor with the sources modal open (because of the `modal=sources` in the URL).

pathmapper:
Thanks @orangemug, some comments:

- For the Maputnik Logo with version in the top left corner the link to `https://github.com/maputnik/editor` is gone (I'm personally using it a lot :-))

- Showing the docs/SDK info was previously possible by clicking on a field name (where the question mark pointer is shown). This isn't working anymore, which is no problem, but I think we don't need to change the pointer to a question mark anymore if it stays like this that you could open the docs only using the info button.

- Export modal: Either "Access Token" should be written entirely in the second line for all token fields like it is currently done for the Thunderforest Access Token, or the whole field name should use only one line for each token.

- How do you unfocus/navigate out of the JSON editor using only keys when you are "inside"? I've used the tab navigation to get inside where the code is, but I didn't found any way to loose focus.

> I now also mention 'color accessibility' in the view selector

Nice!


> do you think we should keep the new URL router?

I like that you could share a URL with a certain modal open. It could be that the layer parameter could confuse some users, because it's not obvious that a part is a style hash value. Also when the style is modified, the hash is constantly updated which brings some "noise" to the URL. (btw, does the hash generation affects the editor performance in any way?)
If we keep the new URL router, I would prefer if we would remove the layer query parameter from the URL. Maybe we could add a setting to the debug modal where the layer query parameter (and thereby the hash generation) could be switched on/off).

pathmapper:
After further thinking about it, I think we should keep the URL router because it's really useful for debugging and the docs use case you mentioned @orangemug. Open question for me:

- Does the hash generation affects the editor performance noticeable (especially for large styles)?

pathmapper:
> How do you unfocus/navigate out of the JSON editor using only keys when you are "inside"? I've used the tab navigation to get inside where the code is, but I didn't found any way to loose focus.

Ok, this is related to https://github.com/maputnik/editor/issues/322.

orangemug:
> Does the hash generation affects the editor performance noticeable (especially for large styles, and e.g. using the sliders for max/min zoom or moving around the color picker)?

No it shouldn't, it's a one way operation on style edit. The style state updates the querystring parameters, it's only on loading/reloading the editor that the URL gets parsed.

> For the Maputnik Logo with version in the top left corner the link to https://github.com/maputnik/editor is gone (I'm personally using it a lot :-))

Fixed

> Showing the docs/SDK info was previously possible by clicking on a field name (where the question mark pointer is shown). This isn't working anymore, which is no problem, but I think we don't need to change the pointer to a question mark anymore if it stays like this that you could open the docs only using the info button.

Fixed

> Export modal: Either "Access Token" should be written entirely in the second line for all token fields like it is currently done for the Thunderforest Access Token, or the whole field name should use only one line for each token.

Fixed

Screenshot 2020-06-30 at 09 48 41


> How do you unfocus/navigate out of the JSON editor using only keys when you are "inside"? I've used the tab navigation to get inside where the code is, but I didn't found any way to loose focus.

Yeah so you can press `ESC` to lose focus, but it's not ideal. That bit is definitely something to make better later on

pathmapper:
Looks good!

pathmapper:
Uh, wait, there's still the issue with the question mark pointer.

pathmapper:
You have to hit exactly the info button to expand the doc section. 

If you hover the name, the pointer changes to a question mark but nothing happen if you click. 

question_mark_pointer

orangemug:
So in the latest commit I've removed pointer from the label and increased the hit target of the info button (added some padding). Does that seem ok @pathmapper ?

pathmapper:
Perfect, thanks @orangemug.

Looks good to merge now :+1:

orangemug:
Thanks, merging!

PR #713, opened 2020-07-22
Add new query params to allow better control of ApiStyleStore

ianbelcher:
I'm currently trying to integrate Maputnik into our platform and haven't been able to find a viable way to allow a non-technical map designer to work on a number of different styles without the need for a technical user level install or an offline, communication based process.

With these proposed changes, a link can be given to a non-technical map designer which allows them to modify a given map design which exists on a standard REST API endpoint that they would require access to.

This method also allows the decoupling of the CLI tool from the front-end which makes integrating the tool easier in more contexts as well. i.e. Services with Kubernetes.

By passing through the query params of `styleApiProtocol`, `styleApiHost`, `styleApiPort` and `styleApiPath` Maputnik can then access any REST API endpoint to get styles. At the moment, ApiStyleStore selects the first available style id, so I have also added the param `styleApiPreselectedId` which allows for the specific selection of style where more than one exist.

As an example:
```
https://maputnik.github.io/?styleApiProtocol=https&styleApiHost=example.com&styleApiPort=80&styleApiPath=%2Fapi%2Fstyles&styleApiPreselectedId=xxxyyyzzz
```

The above link will allow the editing of the style at https://example.com/api/styles/xxxyyyzzz as long as the endpoint implements typical REST measures.

Thought I'd put this forward to gather feedback before doing any further work. Any pointers with how this might be tested in the current suite would be great as well as how and where I should document this feature.

### DIFF
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -95,14 +95,19 @@ export default class App extends React.Component {
 
     this.revisionStore = new RevisionStore()
     const params = new URLSearchParams(window.location.search.substring(1))
-    let port = params.get("localport")
+    // `localport` param is a legacy undocumented feature. Possibly deprecate this in future if not required?
+    let port = params.get("localport") || params.get("styleApiPort")
     if (port == null && (window.location.port != 80 && window.location.port != 443)) {
       port = window.location.port
     }
     this.styleStore = new ApiStyleStore({
       onLocalStyleChange: mapStyle => this.onStyleChanged(mapStyle, {save: false}),
       port: port,
-      host: params.get("localhost")
+      // `localhost` param is a legacy undocumented feature. Possibly deprecate this in future if not required?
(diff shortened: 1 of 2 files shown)


orangemug:
👋 @ianbelcher how about instead we pass in the URL as a whole, so

```
/?styleApi=https://example.com/api/styles
```

You can use the `URL` constructor to parse the URL, so

```js
const urlObj = new URL(params.get("styleApi"));
```

ianbelcher:
@orangemug Will do. Just confirming that you're also happy with the naming of `styleApiPreselectedId`?

Where should this be documented as well?

orangemug:
Sorry for the delay. I'm a little confused what `styleApiPreselectedId` is doing. If you're loading a style from an API can't the thing you're loading from set the id? So you'd pass the following to Maptunik

```
/?styleApi=https://example.com/api/styles/xxxyyyzzz
```

PR #724, opened 2020-09-08
Remove key/value from style when editing style results in empty array

### LINKED ISSUE #707: "Add Value" button is pressed and then the delete button is pressed, an empty array is generated.
**Maputnik version**:v1.7.0

**Description of the bug**:
"Add Value" button is pressed and then the delete button is pressed, an empty array is generated.

For example, perform the above operation with `text-variable-anchor`, `"text-variable-anchor": []`  is generated and displayed letters will disappear.

To revert `text-variable-anchor` to initial state, it cannot be solved by GUI, the only solution is to edit the editor directly.

orangemug:
Fixes #707

See #707 for full description of issue.


pathmapper:
Looks good!

PR #726, opened 2020-09-08
Fix to not change layer key while editing layer

### LINKED ISSUE #718: JSON Editor: Focus is lost when backspace key is used on id value
**Maputnik version**: v1.7.0, master
**Browser**: Chrome
**OS**: Linux

**Description of the bug**:
JSON Editor: Focus is lost when backspace key is used on `id` value

loose_focus

orangemug:
Once we select a layer the components key remains until we select another layer, regardless of whether we edit the layer id in the editor. This fixes the `<LayerEditor/>` component recycling on layer `id` edit, which has the consequences described in #718 

Fixes #718

### DIFF
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -772,7 +772,10 @@ export default class App extends React.Component {
           }
         }
         if (!invalid) {
-          this.setState({selectedLayerIndex});
+          this.setState({
+            selectedLayerIndex,
+            selectedLayerOriginalId: this.state.mapStyle.layers[selectedLayerIndex].id,
+          });
         }
       }
       catch (err) {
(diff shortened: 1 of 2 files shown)


pathmapper:
@orangemug does the codesandbox work for you? I'm getting 

> Invalid Host header

orangemug:
@pathmapper fixed in #728 and updated recently opened PRs with that fix.

pathmapper:
When the layer `id` value is updated in the layer field, the value in the JSON editor is updated only when the focus of the `id` field is lost. Is this expected @orangemug ?

For comparison, the opacity value is updated in the JSON editor while typing in the field:

orangemug:
> Is this expected

@pathmapper you're right this is incorrect, I've fixed it in the latest commit. I think there are probably a few other places that do this also, I think it's a left over from where we didn't have a way have a partially invalid style. We should aim to make this consistent across the UI

PR #729, opened 2020-09-09
Correctly upgrade old-style filter functions to expressions

### LINKED ISSUE #712: $type as key for old syntax filter cause switch to expressions syntax ui and opens export modal
**Maputnik version**: v1.7.0, master
**Browser**: Firefox
**OS**: Linux

**Description of the bug**:

$type as key for old syntax filter cause switch to expressions syntax ui and opens export modal

1. Open the OSM Liberty style
2. Select layer `road_area_pattern`
3. Delete the existing filer using the trash can icon
4. Add a new filter and enter `$type` as key

**Style file or style URL**:
https://maputnik.github.io/editor?style=https://maputnik.github.io/osm-liberty/style.json

**GIF**:
$type

orangemug:
Initial attempt at fixing #712, needs some testing. Feel free to give it a test with the generated demo link below and report any errors.

Fixes #712

### DIFF
--- a/src/components/FilterEditor.jsx
+++ b/src/components/FilterEditor.jsx
@@ -62,24 +62,19 @@ function createStyleFromFilter (filter) {
   };
 }
 
-/**
- * This is doing way more work than we need it to, however validating a whole
- * style if the only thing that's exported from mapbox-gl-style-spec at the
- * moment. Not really an issue though as it take ~0.1ms to calculate.
- */
+const FILTER_OPS = [
+  "all",
+  "any",
+  "none"
+];
+
+// If we convert a filter that is an expression to an expression it'll remain the same in value
 function checkIfSimpleFilter (filter) {
-  if (!filter || !combiningFilterOps.includes(filter[0])) {
-    return false;
+  if (filter.length === 1 && FILTER_OPS.includes(filter[0])) {
+    return true;
   }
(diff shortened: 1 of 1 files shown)


pathmapper:
@orangemug the bug described in #712 is gone :+1:  

The filter value is only updated in the JSON editor when focus is lost (as we discussed before in https://github.com/maputnik/editor/pull/726#issuecomment-689736956) and also the error is only gone after loosing focus. Is this also something which could be change to `onInput`?

orangemug:
> The filter value is only updated in the JSON editor when focus is lost (as we discussed before in #726 (comment)) and also the error is only gone after loosing focus. Is this also something which could be changed to onInput?

@pathmapper hmmm, I might back track on that suggestion a little 😁. So if we do it `onInput` in this instance it would cause errors to appear as you type. This is probably a little annoying because although technically correct, you probably just haven't finished typing yet. I guess the perfect UI would probably be a mix of the both, so `onBlur` and also `onIdle` (I made this one up), so if a user pauses typing for more than 1-2 seconds then the UI would update and the error would show. I think `onBlur` probably makes sense for now, I'll create a little mockup in the next batch of work see if we can get something that 'feels right' without being annoying.

Merge if you're happy with the above ⬆️

Issue #775, opened 2022-03-29
mapbox is outdated


JannikGM:
Maputnik v1.7.0 (as hosted on https://maputnik.github.io/editor) runs using mapbox Version 1.9.1.
https://github.com/maputnik/editor/blob/v1.7.0/package.json#L44

This is an old version of mapbox that doesn't support expressions such as `slice`: https://docs.mapbox.com/mapbox-gl-js/style-spec/expressions/#slice
I'd consider this so limiting, that a new maputnik release should occur after the mapbox upgrade, so these new styling features can be used.

In master, this has been upgraded to 1.11.0 (in https://github.com/maputnik/editor/commit/24327541c5c61389c54484396e404689b48fca92), but that's also outdated by now.
Please consider upgrading mapbox 1.x to the latest 1.13.x + potentially additional support for maplibre in the future.

pathmapper:
With @propheel's PR, `master` is now using mapbox-gl `1.13.2`.

Opened #781 to start a discussion about maplibre support.

sebisteiner:
Any plans on building and deploying a new release?

Issue #762, opened 2022-02-06
Migrate from node-sass to dart-sass


dBitech:
node-sass is now deprecated, as per the following from the authors.

Warning: [LibSass and Node Sass are deprecated](https://sass-lang.com/blog/libsass-is-deprecated). While they will continue to receive maintenance releases indefinitely, there are no plans to add additional features or compatibility with any new CSS or Sass features. Projects that still use it should move onto [Dart Sass](https://sass-lang.com/dart-sass).

PR #776, opened 2022-04-06
Use node 16.x

propheel:
It required converting mocha tests code into async since [@wdio/sync is deprecated](https://webdriver.io/docs/sync-vs-async/) starting with node v16. It removed the dependency on fibers and on [node-gyp + python](https://webdriver.io/docs/sync-vs-async/#common-issues-in-sync-mode) indirectly though which is a great thing.

Also moved away from node-sass to sass since [node-sass is deprecated](https://sass-lang.com/blog/libsass-is-deprecated) which addresses #762 .

It is also a beginning of Maputnik upgrade #775 

Feel free to share comments.


propheel:
Was looking into why test/standalone-chrome keeps failing all the time and apparently there might be [a problem with sandbox + /dev/shm size](https://stackoverflow.com/a/53970825) for Docker image. Will try to use runner machine instead of Docker images.

propheel:
Thank you @pathmapper ! This is actually my first contribution to OSS in my life. I didn't expect such great experience! 😄 

Hope that it will stay along with more PRs that will come 😉

pathmapper:
@propheel glad you chose this project :smile: 

If you would like to tell something about what you're using Maputnik for, please feel free to post it in https://github.com/maputnik/editor/issues/164.

PR #778, opened 2022-04-08
Update dependencies

propheel:
This bumps **mapbox-gl** to **1.13.2** as well as lots of other dependencies like babel, storybook, webpack and webdriverio.

As a result these:
```
added 2345 packages, and audited 2346 packages in 1m
38 vulnerabilities (2 low, 16 moderate, 19 high, 1 critical)
```
came down to these:
```
added 2287 packages, and audited 2288 packages in 60s
24 vulnerabilities (3 moderate, 21 high)
```

It could have been much better though with React 18 and Webpack 5, but [Storybook is not ready yet](https://github.com/storybookjs/storybook/issues/17831).

Still we should have #775 covered.

Also configures selenium so it starts server automatically when running tests without the need to run it manually. Changes are reflected in Readme as well.

Open for comments!


pathmapper:
:+1: 

@propheel I'll have a closer look and do some manual testing.

propheel:
Thank you, sir. I made a number of small decisions along the way that I haven't described, so feel free to comment and I'll also try to describe them.

One of regressions to 1.7.0 I found out is that layers cannot be dragged anymore, but it's because span was changed to button and it is not draggable (sortable?) by default. The fix is a one liner, but it's going to be a part of another PR. Or I can add it to this one if you prefer.

pathmapper:
> I can add it to this one if you prefer.

Looks like this is a regression to `master`, so please add it to this PR.

propheel:
Just added it, @pathmapper 

> this is a regression to `master`

Technically speaking this was introduced [some time ago](https://github.com/maputnik/editor/commit/b28407a4a009225fbad09084f5ec60b1cdccd899#diff-9478961841c4f0dbc795bea1dc17479cddb8e88f92fd3ebfefe87f0858d0bebfL19), so it's not a regression to `master` rather than to `v1.7.0` as I indicated. It's just a thing I noticed when I was doing manual testing.

The fix is to use [`shouldCancelStart`](https://github.com/clauderic/react-sortable-hoc#sortablecontainer-hoc). There you can also find some description why buttons work differently than span.

pathmapper:
Thanks @propheel , but the issue remains - it's not possible to change the layer order by dragging:

layer_drag

pathmapper:
Nice change regarding running the tests :+1: 

For Chrome two tests are failing. Do they pass on your side?

grafik

For Firefox the tests are not running, I get

```
Error: Failed to create session.
[0-0] Could not start a new session. Error while creating session with the driver service. Stopping driver service: Could not start a new session. Response code 400. Message: Argument --remote-debugging-port can't be set via capabilities
```

Any idea how to solve this?

grafik

propheel:
> Thanks @propheel , but the issue remains - it's not possible to change the layer order by dragging:

You're right, @pathmapper . That was a side effect of new array-move lib version adding in-place move. I also fixed initial load warning of selected style.

> Nice change regarding running the tests 👍

Thank you 😊 

> For Chrome two tests are failing. Do they pass on your side?

Yes, however with latest version of Chrome (101.0.4951.41) they get stuck at some point ("resolving host" in browser). Will investigate it. Chrome passes CI though.

> For Firefox the tests are not running, I get
> 
> ```
> Error: Failed to create session.
> [0-0] Could not start a new session. Error while creating session with the driver service. Stopping driver service: Could not start a new session. Response code 400. Message: Argument --remote-debugging-port can't be set via capabilities
> ```
> 
> Any idea how to solve this?

Yup, I can see the same on CI as well. Seems related to https://github.com/webdriverio/webdriverio/issues/8282

pathmapper:
Nice work @propheel, this PR is a huge step forward :tada: 

Thanks for the fixes!

> also fixed initial load warning of selected style

:+1: 

> Yes, however with latest version of Chrome (101.0.4951.41) they get stuck at some point ("resolving host" in browser). Will investigate it. Chrome passes CI though.

Same on my side, since FF is passing locally and CI is passing for FF + Chrome, this should't block us from merging here :-)

Opened #779 to track this is issue.

PR #785, opened 2022-08-12
Add geocoder

ebrelsford:
This change adds a Mapbox GL geocoder to the map controls.

To do:
 * [ ] determine whether we want to always show it, currently we do


src/components/App.jsx
```diff
@@ -687,7 +687,9 @@ export default class App extends React.Component {
         options={this.state.mapboxGlDebugOptions}
         inspectModeEnabled={this.state.mapState === "inspect"}
         highlightedLayer={this.state.mapStyle.layers[this.state.selectedLayerIndex]}
-        onLayerSelect={this.onLayerSelect} />
+        onLayerSelect={this.onLayerSelect}
+        showGeocoder={true}
```
ebrelsford, line 691: Do we always want to show it? Only when there's a valid Mapbox API key?

pathmapper: Hi @ebrelsford, thanks for your PR but I think would should not include the Mapbox geocoder at all because of the Mapbox Geocoding API terms:

> 4.1. In General. You may not use Geocodes: [...] (b) to develop any general
> purpose printed or digital map (of any geographic size)

From https://assets.website-files.com/5d4296d7a839ea49599adba1/613b743ce6c1fabefc460c25_Mapbox%20Service%20Terms%20(2021.09.10).pdf

And developing a map is what Maputnik is for :)

Issue #753, opened 2021-08-18
Undo undoes too much


doskabouter:
**Maputnik version**:v1.7.0
**Browser**: Firefox latest
**OS**:Windows

**Description of the bug**:
If you change multiple things, ctrl-z undoes them all in one go

**Steps to reproduce the behavior**:
1. Open random mapstyle, select one of the items
2. Change something in the filter
3. Change a color
4. press ctrl-z

Expected result would be that at first press, the color is reverted, and on the second press the filter is reverted.
Even selecting a different item inbetween step 2 and 3 (to hopefully trigger a savepoint) results in both actions being undone in one step

Saxteja:
can you tell me how did you redo those changes

doskabouter:
i didn't manage to redo those. and to be honest, I'm not even sure its still an issue

PR #806, opened 2023-03-31
Only store revisions that include style changes

kochis:
Similar to issue https://github.com/maputnik/editor/issues/753, I've run into some odd behavior when trying undo/redo recent changes.

I've been using this change in a local fork, and it's seemed to be running pretty well. The main logic changes are:
* Only store a revision if there is a style diff
  * so you don't need to "undo" no-op changes
* Clear the "redo" revisions once an addition is made
  * without this change, things are _pushed_ to the end of the list, even though the currentIndex may have been decremented (I suspect this is the main bug)

### DIFF
--- a/src/libs/revisions.js
+++ b/src/libs/revisions.js
@@ -1,3 +1,5 @@
+import {diff} from '@mapbox/mapbox-gl-style-spec'
+
 export class RevisionStore {
   constructor(initialRevisions=[]) {
     this.revisions = initialRevisions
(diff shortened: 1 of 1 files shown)


nyurik:
@kochis thx!  Could you give some instructions on testing this? I'm slowly getting back into reviewing it

HarelM:
@kochis I see how this makes sense (assuming style spec diff is trust worthy).
We have now brought everything to a state where this can be maintained.
There's e2e test that checks undo/redo, but they don't check this scenario, so I think it would be good to have either an e2e test or a unit test around this code.
You'll need to migrate this code to use maplibre diff and move it to the relevant typescript file.
Hope this is still relevant for you.

HarelM:
I've tested this code for comments change, which is not defined as a change in terms of the maplibre-style-spec code.
I'm not sure if this is a bug there as comments are not considered to be a style change for maplibre-gl.
The redo fix is probably correct anyway and needs to be added to the code.

HarelM:
Closing this PR for now, I think a proper style change mechanism should be used here but I'm not sure the `diff` of maplibre-style-spec is the right solution.

PR #808, opened 2023-04-21
Use nginx to serve built files

### LINKED ISSUE #842: Reduce docker image size
While not very interesting to solve, it's a very simple change to use nginx instead of python and reduce image size from 50Mb to 5Mb.

nbr23:
Resolves #842
Better suited than python + handles sigterm better


Dockerfile
```diff
@@ -12,11 +12,8 @@ RUN npm run build
...
-FROM python:3-slim
-WORKDIR /maputnik
+# Create a clean nginx-alpine image with just the build results
 
-COPY --from=builder /maputnik/build/build .
+FROM nginx:alpine
 
-EXPOSE 8888
-CMD python -m http.server 8888
+COPY --from=builder /maputnik/build/build /usr/share/nginx/html/
```
nyurik, line 19: I would recommend auto-newline as part of the IDE config - they get annoying if missing when diffing.

Overall, no opinion on this change, if it works, i'm fine with this unless someone else objects

```suggestion
COPY --from=builder /maputnik/build/build /usr/share/nginx/html/

```

nbr23: Indeed, thanks!

HarelM:
Resolves #842, no reason not to add this.
The only difference is that this exposes port 80 while the original image eposes port 8888.
Having said that this is a breaking change version and it makes sense.

PR #810, opened 2023-06-10
Added checkbox for cluster attribute 

Eliav-B:
Added additional checkbox to set the cluster attribute when adding a GeoJson source


nyurik (approved):
looks good, thanks!

PR #826, opened 2023-10-23
Add dependabot

birkskyum:
It has a low limit on how many PRs it's allowed to open at a time, so it should not flood the repo.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,14 @@
+version: 2
+updates:
+  - package-ecosystem: "npm"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    open-pull-requests-limit: 2
+    versioning-strategy: increase
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    open-pull-requests-limit: 2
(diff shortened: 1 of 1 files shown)


nyurik (approved):

HarelM:
In maplibre-gl I've added some flows to automatically merge PRs from dependabot if they pass CI.
It requires setting a branch protection and setting CI runs as required, but it removes a lot of the manual work around upgrading dependencies.

nyurik:
Yeah, we should implement that too.  I think there is tons of work to get this repo into a better state, such as testing, linting, better CI, local testing, updates to dockers, rename primary branch, and zillions of other things ;)

HarelM:
Yup, the best way to solve this it to prioritize I believe.
For me the priorities are:
1. Create an official release including npm + docker (currently there is only a pre-release) - assuming this is semi-automatic, otherwise make is easier to create a new release.
2. Change default branch name to main
3. Automate dependecies upgrade
4. After most of the dependencies are up-to-date (hoping most of them don't have breaking changes) upgrade lint rules
5. Other stuff...

nyurik:
sounds good to me!  Thanks for looking into this!!!

PR #829, opened 2023-12-14
Use driver pattern for e2e tests

HarelM:
I've moved all the logic relevant to WBIO into a single file in order to be able to replace it.
I have tried to upgrade WDIO in order to use the latest version and it got stuck on my computer.
Furthermore, I was not able to run it locally which made this whole cycle very long.
After this will be merged I will replace WDIO with cypress.
This doesn't change anything, only moves some code in the tests to a single file, removes unneeded files and uses the driver pattern, which will later allow switching the underline WDIO with Cypress.

cc: @nyurik


nyurik (approved):
this looks good to me. Minor linting aside (i'm surprised it doesn't get caught by CI), seems like its ready to merge.  Let me know if you want to add any more changes to this, and should be ok

HarelM:
I'll take care of linting later on.
Please merge.

PR #841, opened 2023-12-18
Refactor driver for E2E

ShellyDCMS:
Added shellygo package.
Refactor driver and e2e tests.
Added data-wd attributes to missing places.

PR #843, opened 2023-12-19
Refactor driver for E2E

HarelM:
This is basically the content of #841 with the code review changes and relevant fixes to tests/driver code to pass the tests.
CC: @ShellyDCMS

After this we should lint the project and add the lint to the CI to make sure it doesn't break.

### DIFF
--- a/cypress.config.ts
+++ b/cypress.config.ts
@@ -5,5 +5,9 @@ export default defineConfig({
     setupNodeEvents(on, config) {
       // implement node event listeners here
     },
+    retries: {
+      runMode: 2,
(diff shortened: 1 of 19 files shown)


HarelM:
Hmm...
Seems like "skip link to layer editor" test doesn't want to pass on chrome when running in CI.
When I run it locally it does pass both from CMD and from Cypress UI.
Not sure what to do here...

HarelM:
Merging this to continue development. If the test will continue to be ignored, we'll remove it eventually.
@nyurik Now might be a good time to push the lint fixes

PR #847, opened 2023-12-20
Migrate all the non react components code to typescript

HarelM:
This completes the migration to typescript of all the non react components code.
The only changes introduced besides types are the type checks using `"something" in object` which narrows down types in typescript.


birkskyum (approved):
looks good

PR #848, opened 2023-12-21
Migration of jsx files to tsx 1

HarelM:
In this PR I have changed some of the jsx files to tsx file.
I'm starting off with the "leafs" so that migration of the rest will be easier, hopefully.

What I'm basically doing is taking a jsx file, copy paste it into:
https://mskelton.dev/ratchet

And after that I'm fixing the types as needed.
It's not a very long process.

Hopefully more PRs will follow and this will be over soon.
I don't plan to migrate the storybook as I generally don't understand why is it useful, I'll open an issue to see if anyone thinks differently.


src/components/InputNumber.tsx
```diff
@@ -60,25 +67,25 @@ export default class InputNumber extends React.Component {
     }
 
     this.setState({
-      dirtyValue: newValue === "" ? undefined : newValue,
+      dirtyValue: newValue === "" ? undefined : value,
```
birkskyum, line 70: these changes between value/newValue appear like it could change some logic - is it a result of automatic conversion?

HarelM: More of a typing problem. Value is types correctly and is initialized with newValue in the beginning of the method.
This was not automatic.

birkskyum (approved):

HarelM:
Thanks!! I'll continue to push this forward in another PR to reduce re-reviewing.

PR #850, opened 2023-12-21
Migration of jsx files to tsx 2

HarelM:
This is to continue the work of migrating all the jsx files into tsx files.
The MO is basically described here: #848.

About 7 files to go...


src/components/ModalSources.tsx
```diff
@@ -128,38 +131,38 @@ class AddSource extends React.Component {
...
-        url: source.url || `${protocol}//localhost:3000/tilejson.json`
+        url: (source as VectorSourceSpecification).url || `${protocol}//localhost:3000/tilejson.json`
       }
       case 'tilexyz_vector': return {
         type: 'vector',
-        tiles: source.tiles || [`${protocol}//localhost:3000/{x}/{y}/{z}.pbf`],
-        minZoom: source.minzoom || 0,
-        maxZoom: source.maxzoom || 14
+        tiles: (source as VectorSourceSpecification).tiles || [`${protocol}//localhost:3000/{x}/{y}/{z}.pbf`],
+        minzoom: (source as VectorSourceSpecification).minzoom || 0,
```
HarelM, line 144: I need to check somehow if this is a bug fix or a wrong change...

nyurik (approved):
cursory examination looks good, but i might have missed something

src/components/ModalSources.tsx
HarelM: Seems like this fixes a bug when the default zoom - it is 0 to 22 before the bug fix and 0 to 14 afterwards, so this is a correct change, the value of this is just amazing so early on.

PR #851, opened 2023-12-22
Migration of jsx files to tsx 3

HarelM:
This is in continue to:
- #850
- #848

The last files should be converted as part of this PR, there are only a handful left.


nyurik (approved):
lgtm

HarelM:
That was fast.
Do you prefer that I open a new PR or pile the rest of the files as part of this PR?

nyurik:
its ok if they are very similar to this change to push them here and squash them after

HarelM:
Ok, I'll push them here then, the changes are mainly to add types and rename the files.
It shouldn't take more than a few work hours to finish this.

HarelM:
This is now ready for review.
I have completed the migration.
It's not perfect, but it's a lot better than before.
I also fixed some console errors and warnings.
I'm not an experienced Maputnik user, so this probably needs some more testing before it can be released.

As far as I could tell the release process is by building it in this repo and copying the files to the maputnik.io repo.
It might be a good idea to maybe create an npm package with the build outputs to allow versioning.

Lint and code coverage are probably the next steps, and later on packages upgrade when the coverage is OK, or if someone can define a way to make sure the build is OK manually.

src/libs/apistore.ts
```diff
@@ -1,10 +1,11 @@
 import style from './style.js'
-import {StyleSpecification, format} from '@maplibre/maplibre-gl-style-spec'
+import {format} from '@maplibre/maplibre-gl-style-spec'
+import type {StyleSpecification} from 'maplibre-gl'
 import ReconnectingWebSocket from 'reconnecting-websocket'
 
 export type ApiStyleStoreOptions = {
-  port?: string
-  host?: string
+  port: string | null
```
birkskyum, line 7: Nitpick, but isn't the current syntax using ? cleaner?

birkskyum (approved):

src/libs/apistore.ts
HarelM: It is, but it's also not the same - one is string or null while the other is srting or undefined.
JavaScript is annoying in that aspect...

PR #852, opened 2023-12-25
add code coverage

ShellyDCMS:
Add Cypress code coverage using Istanbul nyc
image


HarelM:
Please add output file to fit ignore (out.json).

HarelM:
Can you grant me permissions to your repo?
I'll fix the package lock and add the codecov CI step.

HarelM:
I have fixed the package lock but it seems that the coverage is causing the test to fail for some reason.
I was able to reproduce this locally in codespaces when running the tests using command line (headless).
I also found this closed issue, which sounds similar, but it looks like it was fixed in latest version, so I don't know:
https://github.com/cypress-io/code-coverage/issues/216

HarelM:
THANKS BTW!

ShellyDCMS:
> I have fixed the package lock but it seems that the coverage is causing the test to fail for some reason. I was able to reproduce this locally in codespaces when running the tests using command line (headless). I also found this closed issue, which sounds similar, but it looks like it was fixed in latest version, so I don't know: [cypress-io/code-coverage#216](https://github.com/cypress-io/code-coverage/issues/216)
> 
> image

Is this still an issue? it seems we just need to add baseUrl

HarelM:
Yes, this is still an issue, you can see it in the Firefox CI run.

ShellyDCMS:
> Yes, this is still an issue, you can see it in the Firefox CI run.

Fixed

.nycrc.json
```diff
@@ -0,0 +1,27 @@
+{
+  "all": true,
+  "extends": "@istanbuljs/nyc-config-typescript",
+  "check-coverage": false,
+  "include": ["src/**/*.ts", "src/**/*.tsx"],
+  "exclude": [
+    "cypress/**/*.*",
```
HarelM, line 7: Can we remove some of the lines here? seems a bit too verbose...

HarelM (approved):
THANKS!

.nycrc.json
HarelM: nit picking, this can be merged with and without this fix.

HarelM:
Nice (besides the low number obviously)!
image

PR #854, opened 2023-12-26
Improve tests

HarelM:
Recover commented out tests, 
Improve usage of helper in driver, 
Added data-wd-key for test to use instead of classes
Moved all non tsx file to a single folder (lib) instead of lib and utils


nyurik (approved):

HarelM:
Thanks @nyurik!! You are working in weird hours... :-D

PR #856, opened 2023-12-30
improve drivers

ShellyDCMS:


cypress/e2e/accessibility.cy.ts
```diff
@@ -1,7 +1,8 @@
+import { then } from "@shellygo/cypress-test-utils/assertable";
```
HarelM, line 1: I'm not sure it's a good thing to know the internal folder structure of the library...

ShellyDCMS: it's not a folder structure, it's an entry point

HarelM: Why not re export it from the index file? This way you can rename the file in the future if you want...

ShellyDCMS: 1. it is separate from the CypressHelper
2. because of the issue with webpack

ShellyDCMS:
@HarelM , can you please push a lockfile? Even when creating a lockfile with codespaces the build fails

HarelM:
My codespaces account is maxed out for this month... :-/
What steps are you doing to generate the lock file in codespaces?
I think the following should work:
1. `nvm install` to get the right node version
2. `npm install`
3. Push the lock file, did you try that?
4. You can also run `npm ci` to check that the lock file is correct.

ShellyDCMS:
@HarelM , 
I delete the lock file,
I delete node modules
I run npm i
Can you try locally maybe?

HarelM:
I did the above with the addition of `npm cache clean --force` to avoid any local cache issues. I also removed the cache part of the build which I don't think is necessary...

cypress/e2e/accessibility.cy.ts
HarelM: This will be fixed using rollup in the lib.

cypress/e2e/cypress-wrapper-driver.ts
```diff
@@ -1,41 +1,19 @@
 import { CypressHelper } from "@shellygo/cypress-test-utils";
 
-export default class CypressWrapperDriver {
+export default class MaputnikCypressHelper {
```
HarelM, line 3: Can you change the file name as well please?

cypress/e2e/maputnik-driver.ts
```diff
@@ -49,125 +109,80 @@ export default class MaputnikDriver {
...
-      this.helper.get.element(selector).clear().type(text, { parseSpecialCharSequences: false });
+      this.helper.get
+        .elementByTestId(selector)
+        .clear()
+        .type(text, { parseSpecialCharSequences: false });
     },
   };
 
   public get = {
+    ...this.helper.get,
```
HarelM, line 164: Wouldn't this break the idea of encapsulating the helping inside the driver?

cypress/e2e/modals.cy.ts
```diff
@@ -1,10 +1,11 @@
...
+import { MaputnikDriver } from "./maputnik-driver";
 
 describe("modals", () => {
-  let { beforeAndAfter, when, get, should } = new MaputnikDriver();
+  let { beforeAndAfter, when, get, then } = new MaputnikDriver();
   beforeAndAfter();
+
   beforeEach(() => {
-    when.setStyle("");
+    when.visit("/");
```
HarelM, line 8: This isn't exactly the same as the original test...

cypress/e2e/modals.cy.ts
```diff
@@ -48,9 +51,9 @@ describe("modals", () => {
       when.click("nav:export");
     });
 
-    it("close", () => {
+    it.skip("close", () => {
```
HarelM, line 54: Is this `skip` intentional?

cypress/e2e/maputnik-driver.ts
```diff
@@ -49,125 +109,80 @@ export default class MaputnikDriver {
...
-        assert.deepEqual(getter(obj), styleObj);
-      });
-    },
-
-    styleStoreEqualToExampleFileData: () => {
-      cy.window().then((win: any) => {
-        const obj = this.get.styleFromWindow(win);
-        this.helper.given.fixture("example-style.json", "file:example-style.json").should("deep.equal", obj);
-      });
+    maputnikStyleFromLocalStorageObj: () => {
```
HarelM, line 172: Is this in use?

HarelM:
Layes tests were a bit hard to review, I hope it's similar as I was not able to fully compare one test to the other.
Generally speaking this looks great!
THANKS!

cypress/e2e/cypress-wrapper-driver.ts
ShellyDCMS: sure

cypress/e2e/modals.cy.ts
ShellyDCMS: We should probably rename setStyle... next PR

cypress/e2e/modals.cy.ts
ShellyDCMS: nope, thanks!

cypress/e2e/maputnik-driver.ts
ShellyDCMS: no. removing it

HarelM:
I really need to find a way to fix this lint rule for the switch case... :-/
I also like 4 spaces and not 2.
out of scope for this PR...

cypress/e2e/layers.cy.ts
```diff
@@ -160,114 +147,138 @@ describe("layers", () => {
...
-                type: "background",
-              },
-            ]
-          );
+        describe("color", () => {
+          let bgId: string;
+          beforeEach(() => {
+            bgId = createBackground();
+            when.click("layer-list-item:background:" + bgId);
+            when.click("spec-field:background-color", 0);
```
HarelM, line 269: Isn't 0 the default? Is this for readability?

cypress/e2e/layers.cy.ts
```diff
@@ -295,13 +306,14 @@ describe("layers", () => {
           when.click("layer-list-item:background:" + bgId);
 
           let errorSelector = ".CodeMirror-lint-marker-error";
-          should.notExist(errorSelector);
+          // Not a good idea to check that something doesn't exist. infinite things don't exist.
```
HarelM, line 309: This checks that the same element does not exist before the click, and does exist after the click.
If you have a better way to test this please suggest, otherwise I believe this comment can be removed as I don't think it's actionable...

HarelM (approved):
Generally approved, THANKS!
I have reviewed and made sure the tests in `layers.cy.ts` are the same.
I have left a few minor comments.

cypress/e2e/layers.cy.ts
ShellyDCMS: Nope. Change the default to undefined, otherwise, all the checks of "non existence" will fail.
In this specific case there are two identical data attributes for different elements. Once this is fixed we can remove the 0 here 
image

HarelM: Is it possible to donit as part of this PR?

ShellyDCMS: Well, I am not very familiar with the code, which one of these does the test intend to use?

HarelM: I've added a commit to resolve this. The `FieldFunction` is a wrapper whereas the `InpucSpec` is a more down the hierarchy, which is what the test needs.
I'm not very familiar with the code either...

PR #858, opened 2024-01-11
Prepare for on boarding

HarelM:
I believe this solves most of the missing checkboxes:
- https://github.com/maplibre/maplibre/issues/352

Let me know if there's anything else missing.
Other configuration can be done after the repo has been migrated I believe.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,5 @@
 The MIT License (MIT)
 
-Copyright (c) 2015 Lukas Martinelli
 Copyright (c) 2024 MapLibre contributors
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,7 @@
 The MIT License (MIT)
 
 Copyright (c) 2015 Lukas Martinelli
+Copyright (c) 2024 MapLibre contributors
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
--- a/README.md
+++ b/README.md
@@ -13,17 +13,13 @@ targeted at developers and map designers.
 
 ## Usage
 
-- :link: Design your maps online at **<https://maputnik.github.io/editor/>** (all in local storage)
+- :link: Design your maps online at **<https://www.maplibre.org/maputnik/>** (all in local storage)
 - :link: Use the [Maputnik CLI](https://github.com/maputnik/editor/wiki/Maputnik-CLI) for local style development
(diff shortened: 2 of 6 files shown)


birkskyum:
I was thinking about having the small [docs / landing page](https://maputnik.github.io/) in this repo as well, and push that to maplibre.org/maputnik. What do you think of that?

birkskyum (approved):

LICENSE
```diff
@@ -1,6 +1,7 @@
 The MIT License (MIT)
 
 Copyright (c) 2015 Lukas Martinelli
+Copyright (c) 2024 MapLibre contributors
```
HarelM, line 4: Should this be only Maplibre or keep it with Lukas as it is now?

birkskyum: Think just maplibre is fine in this case

HarelM:
I was thinking that `maplibre.org/maputnik` will be the editor itself and add some info to the maplibre.org site somewhere, like we do for Martin maybe or in another page.
Maputnik is pretty straight forward, I don't see a reason to add a "full" landing page for it.
But I'm good with everything that we decide.

HarelM:
OK, pushing the button...

LICENSE
wipfli: In all other repos we kept the old copyright holder and just added MapLibre. I think we should do the same here...

nyurik: we can never remove existing copyrights

README.md
```diff
@@ -102,60 +101,10 @@ npm run cy:open
...
+Read more about the MapLibre Sponsorship Program at https://maplibre.org/sponsors/.
 
 ## License
 
-Maputnik is [licensed under MIT](LICENSE) and is Copyright (c) Lukas Martinelli and contributors.
-
-**Disclaimer** This is an independent style editor.
-As contributor please take extra care of not violating any Mapbox trademarks. Do not get inspired by Mapbox Studio and make your own decisions for a good style editor.
+Maputnik is [licensed under MIT](LICENSE) and is Copyright (c) Lukas Martinelli and Maplibre contributors.
+As contributor please take extra care of not violating any Mapbox trademarks. Do not get inspired by other map studios and make your own decisions for a good style editor.
```
nyurik, line 110: inspiration is a bit vague - i am not sure about the legal aspects of that. We must never copy code for sure

nyurik: oh, my appologies, this was in the original text - lets keep it then

HarelM:
This was merged, simply create a PR with the changes you want and let's merge them.

PR #878, opened 2024-02-04
Removes number conversion for dirty input

### LINKED ISSUE #870: Cannot type "0.5" in an Opacity field
**Maputnik version**: main
**Browser**: All
**OS**: All

**Description of the bug**:
It is impossible to type a decimal period in the Opacity field. As soon as `.` is typed, the field value changes to `NaN`, regardless if any digits already typed or the place of the cursor within that number.

1. Open any style, for example, "OSM Liberty"
2. Choose a layer with an Opacity property. For example, "road_motorway"
3. Click on the "Opacity" field and see this:
   image
4. Type `0` and see this:
   image
5. Type '.' and see this:
   image

The same problem exists even when you try to change `55` into `5.5` in hope to change it later to `0.5`

**Workaround**
1. Type `0` in the Opacity field
2. Edit the value in the JSON Editor

HarelM:
Fixes #870

- #870

This regression bug was introduced as part of the refactoring of typescript.
Here:
- #848

I've added tests to cover the scenario so that it won't happen again, hopefully.

### DIFF
--- a/cypress/e2e/layers.cy.ts
+++ b/cypress/e2e/layers.cy.ts
@@ -293,7 +293,7 @@ describe("layers", () => {
             then(get.elementByTestId("spec-field-input:background-opacity")).shouldHaveValue("0.");
           });
 
-          it("should keep revert to a valid value when focus out", () => {
+          it("should revert to a valid value when focus out", () => {
             when.click("layer-list-item:background:" + bgId);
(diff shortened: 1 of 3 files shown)


HarelM:
I'm merging this as this is a critical bug.
@birkskyum @nyurik feel free to add comments here and I'll open a different PR to solve them.

PR #879, opened 2024-02-04
Improve inspect hover UI

### LINKED ISSUE #868: Inspect's popup truncates the property values
**Maputnik version**: v2.0.0-pre.2
**Browser**: All
**OS**: All

**Description of the bug**:
The Inspect's popup does not show much of the properties values. Instead, there is a large gap between the property name and the value.

1. Open any style with a vector source, such as the pre-loaded "OSM Liberty" style
2. Change to Inspect view
3. Move the mouse around the map

**Screenshots**:
image

HarelM:
Fixes #868
- #868

It solves all the block within blocks and all kind of controls that are not required which creates gaps.
I use a simple table, so the width is dynamic but it is always aligned for all the properties and the features.
image

Vary basic stuff.
There's still the layer popup, which I'm not sure I know what it is and might need to be fixed as well.
CC: @zstadler

### DIFF
--- a/src/components/MapMaplibreGlFeaturePropertyPopup.tsx
+++ b/src/components/MapMaplibreGlFeaturePropertyPopup.tsx
@@ -32,10 +32,10 @@ function renderFeature(feature: InspectFeature, idx: number) {
     </div>
     {Object.keys(feature.properties).map(propertyName => {
       const property = feature.properties[propertyName];
-        return <div className="maputnik-popup-layer">
-          <div className="maputnik-popup-feature-left">{propertyName}</div>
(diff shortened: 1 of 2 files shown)


zstadler:
Thanks!

From some experiments I've done, hacking the previous css in the browser, a 30% width for the property name would

- Allow almost all OSM keys to be fully visible
- Provide more room for property values, which tend to be longer

HarelM:
I've updated the code, css etc, it uses a HTML table now which I think is what is the expectations here.
I haven't specified the width, so it's fully dynamic now, the default is 50%, but it will change according to the values in the table.

nyurik (approved):
lgtm, thx!

PR #881, opened 2024-02-08
Readme.md container port changed to 80

glebpinigin:
Activating maputnik with `docker run` requires to bind port 80 from container and not 8888, because nginx is running with default config and listening port 80

For `npm run` port 8888 is explicitly set up [in config](https://github.com/maplibre/maputnik/blob/bc5ecfade6801e16df339a8caba766d28b88de56/vite.config.ts#L8).


HarelM (approved):

HarelM:
Thanks!

Issue #871, opened 2024-02-01
Re-initialize map view after changing style 


zstadler:
**Maputnik version**: main
**Browser**: Chrome, Firefox
**OS**: Windows

**Description of the bug**:
When changing style, the map view may contain features of the previous style

**Steps to reproduce the behavior**:
1. Open Maplibre's [maputnik in this location](https://maplibre.org/maputnik/#13/31.27563/35.12306)
2. Open -> Gallery Styles -> Empty Style
3. Reload the page to be on the safe side
4. Open -> Open URL -> https://raw.githubusercontent.com/IsraelHikingMap/VectorMap/master/Styles/IHM.json
5. The map view would look like this:
   ![image](https://github.com/maplibre/maputnik/assets/1304610/97b37048-a867-4be4-8247-a37d0387c698)
6. Open -> Gallery Styles -> OSM Liberty
7. The map view would look like this, with a mix of both styles:
   ![image](https://github.com/maplibre/maputnik/assets/1304610/99a6af70-446c-4a68-807d-a42d94879e09)
8. Reload the page
9. The map view would look like this, without traces of the previous style:
   ![image](https://github.com/maplibre/maputnik/assets/1304610/9fb39065-1921-43ed-be4a-760ed8efcd65)

HarelM:
Hmm... interesting, might be related to style diff capabilities or limitations.

zstadler:
FWIW, cannot reproduce this with v1.7.0 at https://maputnik.github.io/editor/#13/31.27563/35.12306

HarelM:
Looks like there's wrong handling of style update somewhere in the code.
The following is reported in the console:
![image](https://github.com/maplibre/maputnik/assets/3269297/b63e898a-386a-4b97-b764-67e60ada7a79)

In the past (1.7), the style diff was not able to perform the diff and the map was reloaded with the following warning:
`Unable to perform style diff: Unimplemented: setSprite..  Rebuilding the style from scratch.`
In recent version of maplibre-gl-js there is a more eager attempt to allow style diffing, but it might have bugs.
So this might be a bug in maplibre and not maputnik.
I'll need to see if I can create a minimal reproduction.
Thanks for reporting this issue!

HarelM:
Ok, so I tried to see if I can reproduce this on "plain" maplibre, and I was not able to.
I also looked at the diff code in both maplibre-gl and maplibre-style-spec to see if I see some problems there and it is both covered with tests and runs as expected.
[...]

PR #885, opened 2024-03-09
Temp fix using patch-package for maplibre-inspect issue

HarelM:
This is a temporary fix for:
 - #871

The fix is using `patch-package` to fix the part of the code that is causing the issue, I believe.
Initial tests shows that this is fixing the issue, I hope it is not introducing other issues.
A proper fix and also using the maplibre package will be done once a repo is migrated as can be expected after the following PR is resolved:
- https://github.com/maplibre/maplibre/issues/359

### DIFF
--- a/patches/mapbox-gl-inspect+1.3.1.patch
+++ b/patches/mapbox-gl-inspect+1.3.1.patch
@@ -9,7 +9,7 @@ index ea9bf07..3e7ab38 100644
 -    this.render();
 +    setTimeout(() => {
 +      // The calling method is being called when the sources are being updated.
-+      // render will might call setStyle, which is not a good idea while updating the source.
++      // render might call setStyle, which is not a good idea while updating the source.
 +      // So, we are delaying the render to the next event loop.
 +      this.render();
 +    }, 100);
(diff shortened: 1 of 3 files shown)


patches/mapbox-gl-inspect+1.3.1.patch
```diff
@@ -0,0 +1,18 @@
...
+--- a/node_modules/mapbox-gl-inspect/lib/MapboxInspect.js
++++ b/node_modules/mapbox-gl-inspect/lib/MapboxInspect.js
+@@ -159,7 +159,12 @@ MapboxInspect.prototype._onSourceChange = function () {
+   });
+ 
+   if (!isEqual(previousSources, sources) && Object.keys(sources).length > 0) {
+-    this.render();
++    setTimeout(() => {
++      // The calling method is being called when the sources are being updated.
++      // render will might call setStyle, which is not a good idea while updating the source.
```
nyurik, line 12: will might?

nyurik (approved):
lgtm, but comment needs minor improvement

patches/mapbox-gl-inspect+1.3.1.patch
HarelM: Stupid co-pilot... :-)

Issue #846, opened 2023-12-20
Replace react-aria-menubutton with react-select


birkskyum:
I tried to update react, and react-aria-menubutton complained it isn't compatible with latest react.

We should probably just swap it. Long story short react-aria-menubutton became less relevant in 2021 with the rise of downshift, which itself got overtaken by react-select last few years.

<img width="1338" alt="Screenshot 2023-12-20 at 21 03 44" src="https://github.com/maputnik/editor/assets/74932975/3b33d45d-016d-4f77-b68b-fbcd768d2d83">
<img width="1317" alt="Screenshot 2023-12-20 at 21 04 14" src="https://github.com/maputnik/editor/assets/74932975/7b8b2b4a-8e51-4671-8e4a-98d6dd4a9eec">
<img width="1331" alt="Screenshot 2023-12-20 at 21 04 37" src="https://github.com/maputnik/editor/assets/74932975/1b949da4-bcba-4d59-9c60-549669abb916">

HarelM:
I think the path forward is first to use typescript, otherwise updating dependencies will cause things to break only in run time.
It shouldn't take long to upgrade to typescript.
After that we can swap packages.
I also think there's a need to use a component library instead of the proprietary components written here and the storybook here which I have no clue why we should maintain it.
But it might be just me...
Feel free to review my typescript PRs to streamline the migration to typescript.

birkskyum:
that's a good idea. on that note react-select has proper typing too, but it's a good idea to use a component library

HarelM:
Migration is complete, lint is complete, feel free to push this forward.
The most well known component library for react that might be a good idea to use is [MUI](https://mui.com/).
But I don't have a lot of experience with react to actually say something significant here.

HarelM:
@birkskyum I've created a PR to update react to version 18.
See here:
- https://github.com/maplibre/maputnik/pull/890

Can you tell me what I should check in order to see the issue?

Issue #611, opened 2020-01-30
Switch from react-autocomplete to alternative


orangemug:
'react-autocomplete' (https://github.com/reactjs/react-autocomplete) is no longer maintained and causing warnings

> react-dom.development.js:12357 Warning: componentWillMount has been renamed, and is not recommended for use. See https://fb.me/react-unsafe-component-lifecycles for details.
> 
> * Move code with side effects to componentDidMount, and set initial state in the constructor.
> * Rename componentWillMount to UNSAFE_componentWillMount to suppress this warning in non-strict mode. In React 17.x, only the UNSAFE_ name will work. To rename all deprecated lifecycles to their new names, you can run `npx react-codemod rename-unsafe-lifecycles` in your project source folder.
> 
> Please update the following components: Autocomplete

Tasks

 - Find good currently maintained alternative and migrate to that library. Note it must be WAI-ARIA compliant 

To investigate

 - http://react-autosuggest.js.org/

louwers:
React Autosuggest is also no longer maintained it seems.

PR #890, opened 2024-03-20
Update React to version 18 and other deps

HarelM:
This PR aims at updating some packages.
I'll keep this in draft until I'll be more happy with the results.
Current setup seems to work, I'll let the CI run and see how bad this is.
Packages that needs to be updated/replaces:
- [x] ~autocompete - The only warning left in the console is related to the autocomplete, which probably needs to be updated since it's a package that wasn't updated in the last 6 years.~ #611
- [x] ~Codemirror is also something that will need an update, but it looks too complicated at this point in time, so let's see how this goes.~ #891
- [ ] react-color
- [x] ~react-aria-menubutton~ #846
- [ ] Others?

Most of the changes here are related to types, which is good.


HarelM:
Firefox tests are failing due to the following issue:
https://github.com/cypress-io/cypress/issues/29172

HarelM:
Mostly up-to-date.
There's still a need to update some stale packages, which might prove harder than anticipated due to the fact that there might be a need to use a component library, which is a big effort.
So, this is out of scope for this PR.

nyurik (approved):
lgtm!

nyurik:
is there a reason Firefox fails?  should we keep that test?

HarelM:
See my comment about Firefox.

HarelM:
@birkskyum I need your input here in order to merge this as you commented out about a problem in one of the components...

HarelM:
In #846 you wrote something related to React 18 and incompatibility...

birkskyum:
I think it was just npm that gave deprecation or version mismatch errors.

birkskyum:
I wasn't able to get anything going

HarelM:
Ahh, I see, so this is related to `legacy-peer-deps` flag that I added in order to get around the react version mismatch.
I would like to remove it in the future if possible, but I'm not very optimistic as it requires a lot of effort in replacing these old libraries, and it's not that easy.
In any case, I'll merge this and see what feedback I get from users as the sanity tests that I did seems that things are working as expected and also the tests pass (on chrome) which is good news.

PR #893, opened 2024-03-25
localIdeographFontFamily: false to load CJK fonts

### LINKED ISSUE #892: add toggle `localIdeographFontFamily` options of MapLibre GL JS for CJK fonts
https://maplibre.org/maplibre-gl-js/docs/examples/local-ideographs/

- In MapLibre GL JS, if you need to load CJK fonts, you have to set `localIdeographFontFamily`.
- It seems that current version of Maputnik doesn't provide any feature to toggle this option.
- IMHO, it is better to add anything to toggle this options. What do you think?

Kanahiro:
close #892 

this changes works well for CJK fonts.

Screenshot 2024-03-25 at 22 04 36

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -36,7 +36,7 @@
         "lodash.get": "^4.4.2",
         "lodash.isequal": "^4.5.0",
         "lodash.throttle": "^4.1.1",
-        "maplibre-gl": "^4.1.1",
+        "maplibre-gl": "^4.1.2",
         "maputnik-design": "github:maputnik/design#172b06c",
         "ol": "^6.14.1",
         "ol-mapbox-style": "^7.1.1",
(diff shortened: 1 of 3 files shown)


HarelM:
What would this change do to latin characters? Will they still be using the glyphs?
Also worth opening a PR for maplibre-gl to fix the type of the variable.

Kanahiro:
@HarelM Thank you for checking this PR.
This change seems not to make no change on latin characters for me.

HarelM:
Let's use this PR to update maplibre-gl to avoid introducing "as any", it should be a few days at max.

HarelM:
4.1.2 should be available now.
Let me know if it solves the as any part.

Kanahiro:
@HarelM Thank you for supporting! I've fixed.

HarelM (approved):

HarelM:
Should be published in a few minutes, let me know if you encounter any issue with this.

PR #903, opened 2024-05-21
Update Cypress related packages

HarelM:
Let's see if this adds the relevant information to the CI run details.
CC: @ShellyDCMS


HarelM:
Doesn't look like it added any meaningful information... :-(

HarelM:
Added the terminal reporter according to: https://www.deploysentinel.com/blog/debug-failing-cypress-tests-in-ci#install-terminal-report-plugin

HarelM:
Doesn't look like the reported added more information unfortunately.
It also seems like other tests had failed as part of last commit, so the tests are flaky basically, and it might be that the problem is that and not with firefox... IDK...

HarelM:
Seems like it's more stable now using the latest packages. Let's hope it will stay that way...

PR #908, opened 2024-06-22
Add Versatiles Colorful as example style

louwers:
These are great looking permissively licensed styles maintained by a passionate open source communities. I think they should be included in the example styles. 🙂 

The thumbnails were uploaded to GitHub and should be available pretty much forever, I chose to triple the pixel density because the other thumbnails look bad on my high res monitor.

Screenshot 2024-06-22 at 12 25 56

### DIFF
--- a/src/config/styles.json
+++ b/src/config/styles.json
@@ -1,11 +1,5 @@
 [
   {
-    "id": "Americana",
-    "title": "Americana",
-    "url": "https://zelonewolf.github.io/openstreetmap-americana/style.json",
-    "thumbnail": "https://github.com/maplibre/maputnik/assets/649392/23fa75ad-63e6-43f5-8837-03cdb0428bac"
-  },
-  {
     "id": "dark-matter",
     "title": "Dark Matter",
     "url": "https://cdn.jsdelivr.net/gh/openmaptiles/dark-matter-gl-style@v1.9/style.json",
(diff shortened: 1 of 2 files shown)


louwers:
Looks like we need to be Whitelisted https://github.com/ZeLonewolf/openstreetmap-americana/blob/main/TILE_USAGE.md#tile-usage-policy

I asked on Slack!

HarelM:
The above is talking mainly about the tiles, but the styles might have a different licence. 
Cc: @zelonewolf

src/config/styles.json
```diff
@@ -1,4 +1,10 @@
 [
+  {
```
HarelM, line 2: Can this be sorted alphabetically to avoid favoritism?

louwers: Done.

ZeLonewolf:
Americana style is CC0, so no permission needed or granted.

Tile server has transitioned to OSM US as of the last few weeks, but we haven't updated the text yet.

CC/@quincylvania re approval for white listing

louwers:
I have removed Americana because we cannot be whitelisted at this point.

HarelM:
Can't maplibre.org be whitelisted?

HarelM:
Not sure I fully follow the train of thought. @louwers can you elaborate please?

louwers:
image

No, they recently transferred to OSM US infrastructure, and they are still working on a process.

HarelM (approved):

PR #912, opened 2024-07-12
Update feature id label to $id in feature properties popup

caspg:
Current label, "Feature ID", creates a little bit of confusion. It's not clear that this label isn't really part of the feature properties.

For example, when we want to filter by feature id, we need to use `$id`

```
[==, "$id," 123] 
```


HarelM (approved):

HarelM:
THANKS!

PR #914, opened 2024-07-26
Add links to tile provider authentication docs

ianthetechie:
This probably confused some people in the past, since vector tiles won't even display an access denied image ;)

Before (no information on access keys and where to get them):

image

After (add links to providers in info callout + add a field for Stadia Maps API keys):

image


src/components/ModalExport.tsx
```diff
@@ -113,6 +113,13 @@ export default class ModalExport extends React.Component<ModalExportProps> {
         <p>
           Download a JSON style to your computer.
         </p>
+        <p>
+          <a href="https://docs.maptiler.com/cloud/api/authentication-key/" target="_blank" rel="noreferrer">MapTiler</a>,&nbsp;
+          <a href="https://www.thunderforest.com/docs/apikeys/" target="_blank" rel="noreferrer">ThunderForest</a>,
+          and <a href="https://docs.stadiamaps.com/authentication/" target="_blank" rel="noreferrer">Stadia Maps</a>&nbpsp;
```
ianthetechie, line 119: Trailing `&nbsp;` is because otherwise the following text becomes illegibly squashed.

zstadler:
Should there also be a "Stadia Maps Access Token" field?

ianthetechie:
> Should there also be a "Stadia Maps Access Token" field?

It's not required for most web browser uses; I'll add one either this weekend or Monday.

I see I also missed updating(?) some tests. Will look at that too.

nyurik (approved):

nyurik:
@HarelM any blockers for this one?

HarelM:
Yes, translation is now needed for text.
Also the above comment about adding an input for the token was not addressed.

ianthetechie:
Got side tracked on a few other projects, but got back to adding the token support in the latest edits.

Thanks for letting me know about the translation @HarelM! I didn't know that was a thing 😅 I've also added a few lines explaining that to the README.

src/libs/style.ts
```diff
@@ -55,10 +55,6 @@ function indexOfLayer(layers: LayerSpecification[], layerId: string) {
 }
 
 function getAccessToken(sourceName: string, mapStyle: StyleSpecification, opts: {allowFallback?: boolean}) {
-  if(sourceName === "thunderforest_transport" || sourceName === "thunderforest_outdoors") {
```
HarelM, line 58: Won't this change break stuff?

HarelM: Oh, this was just moved elsewhere, nevermind then...

ianthetechie: Yeah sorry I should have flagged that in advance ;) it just moved to avoid all the inspections in multiple places and simply the helper function.

src/libs/style.ts
```diff
@@ -74,7 +70,24 @@ function replaceSourceAccessToken(mapStyle: StyleSpecification, sourceName: stri
   if(!source) return mapStyle
   if(!("url" in source) || !source.url) return mapStyle
 
-  const accessToken = getAccessToken(sourceName, mapStyle, opts)
+  let authSourceName = sourceName
+  let setAccessToken = (url: string) => url.replace('{key}', accessToken)
```
HarelM, line 74: I would simplify this by defining a local variable called `url` or `sourceUrl` and use it instead of defining a local method which gets override in an if staement.

src/locales/zh/translation.json
```diff
@@ -82,7 +82,7 @@
   "Close modal": "关闭模态框",
   "Debug": "调试",
   "Options": "选项",
-  "<0>Open in OSM</0> — Opens the current view on openstreetmap.org": "在 openstreetmap.org 打开当前视图",
+  "<0>Open in OSM</0> &mdash; Opens the current view on openstreetmap.org": "在 openstreetmap.org 打开当前视图",
```
HarelM, line 85: This looks duplicated...?

HarelM:
Yes, translation was added very recently.
Other than the above, this looks good.
It would be great if you can update the initial post with "before" and "after" screenshots.

src/locales/zh/translation.json
ianthetechie: Yeah I'm not sure what's up with that. The i18n tool generated it but I can only find one string in the project.

ianthetechie: Correct.

Are there any instructions on how to test languages? Strangely, I have set `Accept-Language: zh` in my browser to try testing this, but the entire UI is still English, so I'm not sure the best way to verify if we can drop one string.

HarelM: You can use the dropdown to change languages...
Note that the current language is stored in local storage somewhere, so you need to be in incognito mode to see the language from the browser settings...

ianthetechie: oh heh... that's actually somewhat surprising 😅 Turns out the — rather than the HTML entity is the correct one ;)

Current maplibre.org/maputnik Japanese debug UI:

image

Local version after changing the Japanese translation to use the UTF-8 literal:

image

src/libs/style.ts
ianthetechie: I've refactored based on what I think you're suggesting here. For me it's just different and not simpler, but maybe I missed a simpler way to write it.

HarelM (approved):

src/libs/style.ts
HarelM: Thanks!

src/locales/zh/translation.json
HarelM: Thanks!

HarelM:
Thanks!
Should there be an input for stadia maps api key?

ianthetechie:
Ack.

1. I didn't update the screenshots.
2. I think I didn't add the i18n bits for top-level help text I added (it didn't exist when I added that?

I'll address both momentarily.

ianthetechie:
Hmm... I wonder if it is easier to localize these as separate help text strings... I don't know the grammatical structure of Hebrew, but I imagine there is some language where it might be complicated to do this all correctly.

Thoughts on moving it out of the single header text to the expandable help? Here's what it looks like now:

image

ianthetechie:
Ick; that's going to be fun... Sorry I'm not much of a frontend dev, but it looks the HTML is being escaped if I try specifying strings this way...

image

So it's going to be a bit of work with either approach :/

HarelM:
I think adding a description like other fields is a good idea.

ianthetechie:
Okay. This will take some effort to untangle (for me at least). Even adding a relatively simple "learn more" link along with the help text is proving to be difficult.

image

I was able to get this far with it, but hit two roadblocks:

1. I can't figure out how to get a `t` function down to translate the learn more link text.
2. The link is valid, and it fires an `onClick` handler, but left click doesn't work. I assume something in the hierarchy is preventing propagation.

src/components/ModalExport.tsx
```diff
@@ -130,6 +130,12 @@ class ModalExportInternal extends React.Component<ModalExportInternalProps> {
             value={(this.props.mapStyle.metadata || {} as any)['maputnik:thunderforest_access_token']}
             onChange={this.changeMetadataProperty.bind(this, "maputnik:thunderforest_access_token")}
           />
+          <FieldString
```
HarelM, line 133: I would recommend, instead of using the `FieldString` to maybe use the react components that this filed is built from, this way you'd have more control, and there won't be a need to add something specific to Doc component, which is a genereal one.
This is just a though...

HarelM: In order to pass down `t` you simply need to pass it down to the relevant component, but since doc is part of block which is part of fieldsstring (IIRC) you'll need to pass it down a lot of levels, which is not trivial (and hackish).
Another possibility is translating it in the upper part and passing it down.
You might be able to construct a React Element and pass it down instead of passing down only a string, but that will require some typings updates in order to achieve this, and thay might not be trivial as well.
Let me know if I can help in any way.

ianthetechie: Thanks as always :) I think it's reasonable to improve the doc component with linkable text (and found what was causing the issue), so I just gave that a shot.

ianthetechie:
Took a while for me to get back to this again, but I overcame the event and i18n issues finally ;)

The only thing remaining is getting the new strings translated. I also see the i8n plugin keeps weirdly rewriting the file, so a bit of manual cleanup required for some strings.

src/components/Doc.tsx
```diff
@@ -85,7 +87,12 @@ export default class Doc extends React.Component<DocProps> {
             </table>
           </div>
         }
+        {docUrl && docUrlDescription &&
+          <div className="SpecDoc__learn-more">
+            <a href={docUrl}  target="_blank" rel="noreferrer" onClick={() => console.log('Link clicked')}>{docUrlDescription}</a>
```
HarelM, line 92: Remove console log

src/locales/he/translation.json
```diff
@@ -73,16 +72,16 @@
...
+  "Zoom:": "זום:",
   "Close popup": "סגירת החלון",
   "cursor:": "סמן",
   "center:": "מרכז:",
   "rotation:": "סיבוב:",
   "Close modal": "סגירת חלונית",
   "Debug": "דיבאג",
   "Options": "אפשרויות",
-  "<0>Open in OSM</0> &mdash; Opens the current view on openstreetmap.org": "<0>פתיחה ב-OSM</0> - פתיחה של התצוגה הנוכחית ב- openstreetmap.org",
+  "<0>Open in OSM</0> &mdash; Opens the current view on openstreetmap.org": "__STRING_NOT_TRANSLATED__",
```
HarelM, line 84: I think this shouldn't be here...? It is traslated later on...

src/locales/ja/translation.json
```diff
@@ -73,16 +72,16 @@
...
+  "Zoom:": "ズーム:",
   "Close popup": "ポップアップを閉じる",
   "cursor:": "カーソル",
   "center:": "中央:",
   "rotation:": "回転角度:",
   "Close modal": "モーダルを閉じる",
   "Debug": "デバッグ",
   "Options": "設定",
-  "<0>Open in OSM</0> &mdash; Opens the current view on openstreetmap.org": "現在のビューを <0>openstreetmap.org で開く</0>",
+  "<0>Open in OSM</0> &mdash; Opens the current view on openstreetmap.org": "__STRING_NOT_TRANSLATED__",
```
HarelM, line 84: Same here.
I think there's some bug related to the dash...

HarelM:
Looks good, thanks!
Tagging the traslators:
de	German	@josxha
en	English	@HarelM
fr	French	@lhapaipai
hr	Hebrew	@HarelM
ja	Japanese	@keichan34
zh	Simplified Chinese	@jieme

src/components/Doc.tsx
```diff
@@ -18,14 +18,16 @@ type DocProps = {
     'sdk-support'?: {
       [key: string]: typeof headers
     }
+    docUrl?: string,
+    docUrlDescription?: string
```
HarelM, line 22: I think docUrlTitle or docUrlText might be a better name...? Description was confusion for me...

src/locales/de/translation.json
```diff
@@ -176,8 +176,11 @@
   "Error:": "Fehler:",
   "MapTiler Access Token": "MapTiler Zugriffstoken",
   "Public access token for MapTiler Cloud.": "Öffentlicher Zugriffstoken für MapTiler Cloud.",
+  "Learn More": "__STRING_NOT_TRANSLATED__",
```
josxha, line 179:
```suggestion
  "Learn More": "Mehr erfahren",
```

src/locales/he/translation.json
```diff
@@ -177,8 +176,13 @@
   "Error:": "שגיאה",
   "MapTiler Access Token": "MapTiler Access Token",
   "Public access token for MapTiler Cloud.": "Public access token for MapTiler Cloud.",
+  "Learn More": "__STRING_NOT_TRANSLATED__",
```
HarelM, line 179:
```suggestion
  "Learn More": "מידע נוסף",
```

src/locales/de/translation.json
```diff
@@ -176,8 +176,11 @@
...
   "Thunderforest Access Token": "Thunderforest Zugriffstoken",
   "Public access token for Thunderforest services.": "Öffentlicher Zugriffstoken für Thunderforest-Dienste.",
+  "Stadia Maps API Key": "__STRING_NOT_TRANSLATED__",
+  "API key for Stadia Maps.": "__STRING_NOT_TRANSLATED__",
```
josxha, line 183:
```suggestion
  "Stadia Maps API Key": "Stadia Maps API-Schlüssel",
  "API key for Stadia Maps.": "API-Schlüssel für Stadia Maps.",
```

src/locales/he/translation.json
```diff
@@ -73,16 +72,16 @@
   "Collapse": "הקטנה",
   "Expand": "הגדלה",
   "Add Layer": "הוספת שכבה",
-  "Zoom:": "זום:",
   "Search": "חיפוש",
+  "Zoom:": "זום:",
```
HarelM, line 76: Why was this moved? I would consider sorting the files alphabetically - keywise...

src/locales/fr/translation.json
```diff
@@ -177,10 +176,14 @@
   "Error:": "Erreur :",
   "MapTiler Access Token": "Jeton d'accès MapTiler",
   "Public access token for MapTiler Cloud.": "Jeton d'accès public pour MapTiler Cloud.",
+  "Learn More": "__STRING_NOT_TRANSLATED__",
```
lhapaipai, line 179:
```json
{
  "Learn More": "En savoir plus",
  "Stadia Maps API Key": "Clé d'API Stadia Maps",
  "API key for Stadia Maps.": "Clé d'API pour Stadia Maps."
}
```

src/locales/he/translation.json
josxha: The code generation `npm run i18n:refresh` orders the entries like this.

ianthetechie: Yeah exactly. Idk... I would kinda expect consistent ordering across machines but maybe it doesn't. I just ran the refresh command; no manual edits.

src/locales/fr/translation.json
lhapaipai, line 179:
```suggestion
  "Learn More": "En savoir plus",
```

src/locales/fr/translation.json
```diff
@@ -177,10 +176,14 @@
...
   "Thunderforest Access Token": "Jeton d'accès Thunderforest",
   "Public access token for Thunderforest services.": "Jeton d'accès public pour les services Thunderforest.",
+  "Stadia Maps API Key": "__STRING_NOT_TRANSLATED__",
```
lhapaipai, line 182:
```suggestion
  "Stadia Maps API Key": "Clé d'API Stadia Maps",
```

src/locales/fr/translation.json
```diff
@@ -177,10 +176,14 @@
...
+  "API key for Stadia Maps.": "__STRING_NOT_TRANSLATED__",
```
lhapaipai, line 183:
```suggestion
  "API key for Stadia Maps.": "Clé d'API pour Stadia Maps.",
```

src/components/Doc.tsx
ianthetechie: Ahh, good call. Can make that better.

Or would a nested object be even clearer, since they are kindof a pair?

src/components/Doc.tsx
```diff
@@ -18,14 +18,16 @@ type DocProps = {
...
   }
-};
+};  // & WithTranslation?
```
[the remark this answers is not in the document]
ianthetechie: hmm wat... I think that was a note to self in development that wasn't supposed to be committed.

src/locales/ja/translation.json
ianthetechie: Yeah, I tried a few variations and was unsuccessful looking for a solution. It does indeed seem related to the dash.

src/components/Doc.tsx
ianthetechie: Fixed.

src/locales/ja/translation.json
ianthetechie: image

Screenshot after removing this line.

ianthetechie:
Ok... I _think_ the only outstanding item now are translations from @keichan34 @jieme

src/components/Doc.tsx
HarelM: I think it's OK as it is now.

ianthetechie: Sounds good. I went with `docUrlLinkText` for better clarity on how it's used.

HarelM:
We'll give it a few days and then decide if we merge as is or do something else.

src/locales/ja/translation.json
```diff
@@ -177,10 +175,15 @@
   "Error:": "エラー:",
   "MapTiler Access Token": "MapTiler アクセストークン",
   "Public access token for MapTiler Cloud.": "MapTiler Cloud の公開用アクセストークン",
+  "Learn More": "__STRING_NOT_TRANSLATED__",
```
keichan34, line 178:
```suggestion
  "Learn More": "詳細はこちら",
```

HarelM:
I'll merge it like that, if people are interested in seeing Maputnik in their language, they can contribute the relevant translation I guess...

PR #917, opened 2024-08-08
Add react-i18next for multi-language support

### LINKED ISSUE #746: multi-language
We have added a switchover between Chinese and English versions to version 1.7, can we consider adding it in the next version?

keichan34:
This is a rough start on adding react-i18next. I'll be working on adding more translatable strings and translations in the coming days. I'm going to need to wrap class components in HOCs, so let me know if there's something I should be fixing before doing that. I'm thinking now to keep the exported class names exactly the same, and rename the existing classes by prefixing an `I` (for internal). For example:

```
export default class AppToolbar ...
```

becomes

```
class IAppToolbar ...
const AppToolbar = withTranslation()(IAppToolbar);
export default AppToolbar;
```

I'll be able to contribute Japanese strings (I've talked to a couple people on my team and they'll be happy to help as well), so that's the language I decided to go with in this PR.

Closes #746


HarelM:
The main pain point right now is the warning in the console for the auto complete component. If you can fix that before/during introduction of i18n it would be great. 
Also, if there's a component library that can be used that supports i18n out of the box it might be a direction worth exploring.

keichan34:
@HarelM Thanks. I'll check out the auto complete component while I'm working on this, but doing a cursory search, it looks like there might not be much to translate in the auto complete component? I think it's just being used for values in the style itself -- not for field labels, etc? Unless I'm misunderstanding something?

HarelM:
It's less about the i18n of the auto complete component, it more about replacing it with something that doesn't create console warning.
I get why it might not be relevant at all for this PR though.
I'm not a react expert, so I was hoping someone like you could easily fix this.

i18next-parser.config.js
```diff
@@ -0,0 +1,19 @@
+export default {
```
HarelM, line 1: Can we use a typescript file instead of javascript?

src/components/AppToolbar.tsx
```diff
@@ -95,7 +97,7 @@ type AppToolbarProps = {
   renderer?: string
 };
 
-export default class AppToolbar extends React.Component<AppToolbarProps> {
+class IAppToolbar extends React.Component<AppToolbarProps & WithTranslation> {
```
HarelM, line 100: Can `& WithTranslation` be added to `AppToolbarProps` instead?

src/components/AppToolbar.tsx
```diff
@@ -250,12 +257,35 @@ export default class AppToolbar extends React.Component<AppToolbarProps> {
             </label>
           </ToolbarSelect>
 
+          <ToolbarSelect wdKey="nav:inspect">
```
HarelM, line 260:
```suggestion
          <ToolbarSelect wdKey="nav:language">
```

src/components/AppToolbar.tsx
HarelM: If this is internal as you described in the comment, I would call it `AppToolbarInternal`, the prefix for `I` is usually used for interface and it causes an unneeded confusion.

src/components/FieldComment.tsx
```diff
@@ -2,21 +2,23 @@ import React from 'react'
...
+import { WithTranslation, withTranslation } from 'react-i18next';
 
 type FieldCommentProps = {
   value?: string
   onChange(value: string | undefined): unknown
   error: {message: string}
 };
 
-export default class FieldComment extends React.Component<FieldCommentProps> {
+class IFieldComment extends React.Component<FieldCommentProps & WithTranslation> {
```
HarelM, line 13: Same as previous comment.

src/components/FieldComment.tsx
```diff
@@ -2,21 +2,23 @@ import React from 'react'
...
   render() {
+    const t = this.props.t;
     const fieldSpec = {
-      doc: "Comments for the current layer. This is non-standard and not in the spec."
+      doc: t("Comments for the current layer. This is non-standard and not in the spec."),
```
HarelM, line 17: I think it would be better to use a shorter key for translation?

src/components/FieldId.tsx
```diff
@@ -11,9 +12,11 @@ type FieldIdProps = {
   error?: {message: string}
 };
 
-export default class FieldId extends React.Component<FieldIdProps> {
+class IFieldId extends React.Component<FieldIdProps & WithTranslation> {
   render() {
-    return <Block label={"ID"} fieldSpec={latest.layer.id}
+    const t = this.props.t;
+    return <Block label={t("ID")} fieldSpec={latest.layer.id}
```
HarelM, line 18: I'm not sure ID is something we want to translate...

src/components/FilterEditor.tsx
```diff
@@ -155,17 +156,17 @@ export default class FilterEditor extends React.Component<FilterEditorProps, Fil
     })
   }
 
-  static getDerivedStateFromProps(props: FilterEditorProps, currentState: FilterEditorState) {
+  public static getDerivedStateFromProps: GetDerivedStateFromProps<FilterEditorProps & WithTranslation, FilterEditorState> = (props, state) => {
```
HarelM, line 159: Why was this converted to an arrow function?

src/components/FilterEditor.tsx
```diff
@@ -299,17 +304,22 @@ export default class FilterEditor extends React.Component<FilterEditorProps, Fil
...
           {this.state.valueIsSimpleFilter &&
             <div className="maputnik-expr-infobox">
-              You&apos;ve entered a old style filter,{' '}
-              <button
-                onClick={this.makeFilter}
-                className="maputnik-expr-infobox__button"
-              >
-                switch to filter editor
-              </button>
+              <Trans t={t}>
```
HarelM, line 307: Is there a better solution to this? Can we split the relevant translation into 2 keys?

src/components/LayerList.tsx
```diff
@@ -310,6 +317,9 @@ class LayerListContainer extends React.Component<LayerListContainerProps, LayerL
   }
 }
 
+// eslint-disable-next-line react-refresh/only-export-components
```
HarelM, line 320: Would be great if you could find a way to fix these lint issues, I didn't find a better way to do it beside ignore the lint issue, but it's not a great solution.

HarelM:
Thanks for taking the time to open this PR!
Can you add a screenshot of the language button?
Can you also add a e2e test to check language switch and make sure the text is changed?

HarelM:
Also I'm not sure what should happen for RTL languages - i.e. translating to Hebrew and Arabic requires more than simple string substitution, and that's why I asked about using a component library that supports i18n, or something similar.
IDK...

src/components/FilterEditor.tsx
keichan34: Ah, sorry, I was getting a type error when passing the class to `withTranslation`. I think I can fix it another way. I think it was complaining that the props weren't `Readonly<T>`, so I just defaulted to importing the `GetDerivedStateFromProps` type from React. I'll double check.

src/components/FilterEditor.tsx
keichan34: I'd rather not split translations unless completely necessary -- you brought up the point of RTL, but not only that, depending on the language it might be more natural to have elements in a different order than English. I haven't committed these changes yet, but for example, here:

https://github.com/maplibre/maputnik/blob/35840409b8435bc7a68502a6402cce8219b6f9ed/src/components/InputJson.tsx#L167

The translation in Japanese would make more sense as `<kbd>ESC</kbd> ..text..`

keichan34:
I haven't had much experience doing translation in RTL languages, if you know anyone who uses RTL and can help us out, that would be great. I think really supporting RTL would be a bit more involved than just replacing strings, like you said -- we'd probably have to change the layout of the map to be on the left, the toolbar, etc, so there would probably be some CSS and manual `i18next.dir() === 'rtl' ? ... : ... ` usage here and there...

Will be adding E2E tests. Here's what the language button/menu looks like now.

Screenshot 2024-08-14 at 16 36 03
Screenshot 2024-08-14 at 16 36 11
Screenshot 2024-08-14 at 16 36 21

Let me know if maybe it should go somewhere else. I just put it here because it was easy to copy-and-paste the view menu.

Thanks for taking the time to take a look at the PR.

src/components/FieldComment.tsx
keichan34: For simplicity and maintainability, I'm using English as the default language. It's possible to use arbitrary keys, but the downsides outweigh the benefits, I think. The biggest benefits of using English as the translation key, I think, are: 
* familiarity for development (right now, I think most developers working on this project are mainly focused on English-based audiences)
* being able to pick up on changes - for example, if we use a different key structure (in this case, something like `fieldcomment.comment.docstring`), and a developer updates the meaning of this string, they'll only update the translation file of the language(s) that they know. It's easier to find a new string that hasn't been translated yet than it is to find a string that had its meaning changed. This can be prevented by making rules like "if you change the meaning of a string, it needs to be a new translation ID", but I think at this point, translation should be as unobtrusive to the people who are familiar with the codebase as possible.

src/components/AppToolbar.tsx
keichan34: Got it. I'll rename everything to `~Internal` for the internal components

src/components/FieldComment.tsx
HarelM: Fair enough. thanks.

src/components/FilterEditor.tsx
HarelM: I would prefer to have it only when absolutely needed, and I'm not sure this is the case with this text and button.
See the docs about trans - they are implying that in most cases, it's not needed:
https://react.i18next.com/latest/trans-component

HarelM:
I'm speaking Hebrew so I can help out in terms of RTL.
It does require more work by adding the `dir` attribute to all kind of places in the code and have this configured globally somehow.
I have an app which is uses angular material and is fully RTL and LTR compatible here: https://israelhiking.osm.org.il/ but the extra `dir` is not an easy fix.
It would be nice however to include the ability to add `dir=...` as part of this PR.

keichan34:
@HarelM Took a look at isrealhiking. I see. I did some initial RTL support, can you take a look at it and see if the general direction is OK? It looks like if flexbox is used for layout, most LTR->RTL transformations are "free" by just setting `document.body.dir = "rtl"`. The main layout was using position: absolute, so I refactored it to use display: flex, and it looks like it's working, although there may be some minor problems. I'll be working on them in the coming days.

Screenshot 2024-08-15 at 18 36 42

I'm not familiar with RTL, so let me know if there should be any more major layout fixes. Maybe the carets to expand sections and the labels of the sections / layer names should be reversed as well?

src/components/AppLayout.tsx
```diff
@@ -23,17 +24,21 @@ class AppLayout extends React.Component<AppLayoutProps> {
   }
 
   render() {
+    document.body.dir = this.props.i18n.dir();
```
HarelM, line 27: This is great, but I think there are places where we want to "force" `dir=ltr`, for example in the json viewer and in place that we don't translate, like the layer properties names, I think.

src/components/FieldFunction.tsx
```diff
@@ -370,7 +370,7 @@ export default class FieldFunction extends React.Component<FieldFunctionProps, F
     else if (dataType === "data_function") {
       // TODO: Rename to FieldFunction **this file** shouldn't be called that
       specField = (
-        <DataProperty
+        <DataPropertyInternal
```
HarelM, line 373: Is this correct? I wouldn't expect to see this, moreover, I think it should not be exported.

src/components/InputDynamicArray.tsx
```diff
@@ -62,8 +64,12 @@ export default class FieldDynamicArray extends React.Component<FieldDynamicArray
   }
 
   render() {
+    const t = this.props.t;
     const inputs = this.values.map((v, i) => {
-      const deleteValueBtn= <DeleteValueInputButton onClick={this.deleteValue.bind(this, i)} />
+      const deleteValueBtn= <DeleteValueInputButton 
+        onClick={this.deleteValue.bind(this, i)}
+        {...{t, i18n: this.props.i18n, tReady: this.props.tReady}}
```
HarelM, line 71: What's the use of the `{...{`? Is there a simpler way to write this?

src/components/ModalSourcesTypeEditor.tsx
```diff
@@ -167,20 +179,27 @@ class VideoSourceEditor extends React.Component<VideoSourceEditorProps> {
       });
     }
 
+    const cornerLabels: { label: string, key: string }[] = [
```
HarelM, line 182: This looks like a duplicate of the previous one, maybe there's a way to move it to a single place and use it in both places?

i18next-parser.config.ts
```diff
@@ -0,0 +1,19 @@
...
+
+  // We use plain English keys, so we disable key and namespace separators.
+  keySeparator: false,
+  namespaceSeparator: false,
+
+  defaultValue: (locale, ns, key) => {
+    if (locale === 'en') {
+      return key;
+    }
+    return '__STRING_NOT_TRANSLATED__';
```
HarelM, line 17: I would keep the English translation as default.

src/locales/he/translation.json
```diff
@@ -0,0 +1,198 @@
...
+  "Move layer down": "הורד שכבה",
+  "Layer: {{layerId}}": "שכבה: {{layerId}}",
+  "Layers": "שכבות",
+  "Collapse": "הקטנה",
+  "Expand": "הגדלה",
+  "Add Layer": "הוספת שכבה",
+  "Zoom:": "זום:",
+  "Search": "חיפוש",
+  "Close popup": "סגירת החלון",
+  "cursor:": "סמן"",
```
HarelM, line 87:
```suggestion
  "cursor:": "סמן",
```

src/locales/he/translation.json
```diff
@@ -0,0 +1,198 @@
...
+  "Glyphs URL": "כתובת סמלילים",
+  "Center": "מרכז",
+  "Bearing": "כיוון",
+  "Pitch": "זווית",
+  "Light anchor": "עוגן אור",
+  "Light color": "צבע האור",
+  "Light intensity": "חוזק האור",
+  "Light position": "מיקום האור",
+  "Terrain source": "מקור terrain",
+  "Terrain exaggeration": "כמה "להגזים"",
```
HarelM, line 125:
```suggestion
  "Terrain exaggeration": "כמה \"להגזים\"",
```

HarelM:
I've added Hebrew translation, please remove the survey translations and the relevant component as it is no longer in use.

src/components/FieldType.tsx
```diff
@@ -32,15 +34,15 @@ export default class FieldType extends React.Component<FieldTypeProps> {
...
-            ['background', 'Background'],
-            ['fill', 'Fill'],
-            ['line', 'Line'],
-            ['symbol', 'Symbol'],
-            ['raster', 'Raster'],
-            ['circle', 'Circle'],
-            ['fill-extrusion', 'Fill Extrusion'],
-            ['hillshade', 'Hillshade'],
-            ['heatmap', 'Heatmap'],
+            ['background', t('Background')],
```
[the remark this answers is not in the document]
keichan34: I'll revert these style spec translations: I eventually want to make the style spec documentation translatable as well, I think eventually when those are translated, we can pull translations from the style spec itself

src/components/InputDynamicArray.tsx
keichan34: `{...` is basically React's syntax for "the rest of the props for this component should come from this object.

The alternative would be to A) specify each property manually (`t={t} i18n={this.props.i18n}...`), B) create a `i18nProps` variable like `const i18nProps = {t, i18n: .., tReady: ..}; ...<component {...i18nProps}>`, or C) wrap each component in a HOC. If we were using function components, I wouldn't have a problem using the `useTranslation` hook in each component, but HOCs are a bit unwieldily, so I decided to just passthrough the translation props from the parent.

I'll clean it up by making `i18nProps` (B) for now; let me know if there's a better way to clean it up

HarelM:
Thanks for taking care of the Firefox annoying test failures.
If they continue to fail I'll simply remove them from CI.
Do you want this merged or are you planning more changes?

i18next-parser.config.ts
HarelM: I was thinking about returning the key for every language in case the string is not found, and maybe calling a console.warn or something.
Your call, I'm good with either solution you decide.

keichan34: I'll do a console.warn

HarelM:
Can you also delete the following file and remove references to it?
`ModalSurvey.tsx`
Thanks!

keichan34:
Let me know if there is anything else I should brush up.

HarelM (approved):

HarelM:
Let's push it and see how it behaves :-)

HarelM:
Looks nice! A lot better than I expected!
image

I think "zoom:" is missing in the zoom component.
Search placeholder for the geocoder control now updating it with typescript, so it should be easier soon)

We might want to add a test to see that all the keys exist in the translation file, it should be a relatively simple test.

HarelM:
P.S. the word Language should be kept in english, otherwise if you accidentally change the language, it's hard to understand where to click in order to revert to English.
"שפה" in the above screenshot.

keichan34:
Ah: I forgot. The maplibre controls only pick up the translation on initialization, so if you refresh the page, it should work. It has to do with the controls being outside of React's area of control, so I'll have to listen for the i18next events to make switching working with them. I think we should make a tracking issue? or individual issues? for fixes like these? What do you think?

HarelM:
Sure, tracking issues should be better. If the map controls are hard to update I wouldn't stress about them, your call.

PR #930, opened 2024-09-01
Add translation process

HarelM:
## Launch Checklist

This adds information about how to translate Maputnik and how to get in touch with people who have translated in the past.
@keichan34, @jieme, @lhapaipai - I need your approval for this so that people in the future will know to contact you in case text is being added and they need help translating it.

CC: @ianthetechie

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,8 @@ Maputnik is written in typescript and is using [React](https://github.com/facebo
 
 We ensure building and developing Maputnik works with the [current active LTS Node.js version and above](https://github.com/nodejs/Release#release-schedule).
 
+Check out our [Internationalization guide](./src/locales/README.md) for UI text related changes. 
+
 ### Getting Involved
 Join the #maplibre or #maputnik slack channel at OSMUS: get an invite at https://slack.openstreetmap.us/ Read the the below guide in order to get familiar with how we do things around here.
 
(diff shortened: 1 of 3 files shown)


lhapaipai:
It's ok for me.

src/components/MapMaplibreGl.tsx
```diff
@@ -87,7 +90,12 @@ class MapMaplibreGlInternal extends React.Component<MapMaplibreGlInternalProps,
     this.state = {
       map: null,
       inspect: null,
+      geocoder: null,
+      zoomControl: null,
     }
+    i18next.on('languageChanged', () => {
```
HarelM, line 96: I'm not sure this is the right solution, but I couldn't find a simpler one.
The `shouldComponentUpdate` method is returning false when chaning the language and the `render` method is not called.
I found this solution here:
https://github.com/i18next/react-i18next/issues/831

I'm not sure this is the right solution though...

keichan34: I think this is OK for now. When implementing the i18next, there were some times where I thought that it would have been less complicated to use functional components instead of class components -- things like these will probably be cleaned up when migrating to functional components

keichan34 (approved):
Thank you for creating locales/README.md!

ianthetechie:
Thanks for the mention @HarelM! I can strike the less complete localization guide from my work in progress PR now :)

HarelM:
Let's have this deployed, other issues can be fixed later on.

PR #932, opened 2024-09-03
Use same version number for web and desktop

kevinschaul:
With the new release process in place, I think keeping a separate version number for desktop is confusing and unnecessary. This PR remove it so the desktop version reports the version number of the web version embedded in the binary.

## Launch Checklist


HarelM (approved):
Sounds right to me.

HarelM:
I know that in windows you can check excitable file properties and see a version number. Is this number shown there?

kevinschaul:
@HarelM that's a good question. I do not think the version number would be included there currently. A quick search reveals a few tools to embed that information, e.g. https://github.com/tc-hib/go-winres. But unfortunately I do not have access to a Windows machine, so I would be unable to test this.

HarelM:
I can check if you upload it, if that helps.

kevinschaul:
@HarelM Can you check this artifact on Windows? https://github.com/maplibre/maputnik/actions/runs/10707430874?pr=932

HarelM:
Looks good! the only comment I have is regarding the icon (top left corner in the following screenshot), would be nice if it had the Maputnik icon for it instead of the default windows exe icon.
As far as I can tell this can be defined here: https://github.com/tc-hib/go-winres?tab=readme-ov-file#icon-json
image
Overall, this is approved.

kevinschaul:
This may have done it. https://github.com/maplibre/maputnik/actions/runs/10707828481?pr=932

HarelM:
This looks great! Thanks!
image
It might benefit from a larger icon maybe as when you increase it wildly it looks like so (see the PDF for comparison):
image

Another side note is that the following is the desktop experience basically:
image

I was under the impression that this will be more of an electron type of desktop application.
But since this is not the case, I would suggest opening the browser if possible in the given address at least.

In any case all of those are in the nice to have realm, so let me know if you want me to merge this and if you would like to be added as a contributor to this repo.

Thanks for all the work around this!

kevinschaul:
Interesting, I think I'll leave the icon as-is. Good nice-to-have for the future, though. And agreed that it should open the browser. I'll add that as a new issue and hopefully get to it soon! Thank you!

PR #944, opened 2024-09-29
Fix network issue

### LINKED ISSUE #935: Crash - TypeError: i.map is not a function
**Maputnik version**: current https://maplibre.org/maputnik/
**Browser**: Chrome/Firefox
**OS**: Linux

**Description of the bug**:

Maputnik crashes when selecting a layer (maybe related to not available fonts). 

image

1. Load style [gris.json](https://github.com/user-attachments/files/16940152/gris.json) (from https://github.com/qgis/QGIS/issues/58364)
2. Click on the following layer

HarelM:
## Launch Checklist

In case of non 200 response the font and glyphs metadata would return the ajax error object instead of the default value.

Fixes #935

- #935

 - [x] Link to related issues.


pathmapper (commented):
Thanks, confirmed this fixes the issue.

src/libs/metadata.ts
```diff
@@ -5,14 +5,17 @@ function loadJSON(url: string, defaultValue: any, cb: (...args: any[]) => void)
...
+      }
       return response.json();
     })
-    .then(function(body) {
+    .then((body) =>{
       cb(body)
     })
-    .catch(function() {
-      console.warn('Can not metadata for ' + url)
+    .catch(() =>{
```
HarelM, line 17:
```suggestion
    .catch(() => {
```

nyurik (approved):
LGTM.

one side question - can't we use a much more readable await syntax?

```js
try {
  let response = await fetch(url, {
    mode: 'cors',
    credentials: "same-origin"
  });
  if (!response.ok) {
    throw new Error('Failed to load metadata for ' + url);
  }
  let body = await response.json();
  cb(body)
}
catch (ex) {
  console.warn('Can not load metadata for ' + url + ', using default value ' + defaultValue);
  cb(defaultValue)
}
```

HarelM:
In general, this callback hell should be removed, much like we did for maplibre, but it's not a small change...

PR #959, opened 2024-11-24
Gallery Styles: Load different style based on language

louwers:
- Update Protomaps Light to v4.
- Allow loading different (default) style based on selected language. It will replace `$lang` in the style URL with the selected language.
- E2E test with Cypress that confirms the correct style is loaded depended on selected language. 

Screenshot 2024-11-24 at 16 37 25

## Launch Checklist

 - [x] Link to related issues. N/A

### DIFF
--- a/cypress/e2e/i18n.cy.ts
+++ b/cypress/e2e/i18n.cy.ts
@@ -1,5 +1,7 @@
 import { MaputnikDriver } from "./maputnik-driver";
 
+const selectorId = "maputnik-lang-select";
+
 describe("i18n", () => {
   let { beforeAndAfter, get, when, then } = new MaputnikDriver();
   beforeAndAfter();
@@ -8,13 +10,13 @@ describe("i18n", () => {
     it("English", () => {
       const url = "?lng=en";
       when.visit(url);
-      then(get.elementByTestId("maputnik-lang-select")).shouldHaveValue("en");
+      then(get.elementByTestId(selectorId)).shouldHaveValue("en");
     });
 
     it("Japanese", () => {
       const url = "?lng=ja";
       when.visit(url);
-      then(get.elementByTestId("maputnik-lang-select")).shouldHaveValue("ja");
+      then(get.elementByTestId(selectorId)).shouldHaveValue("ja");
     });
   });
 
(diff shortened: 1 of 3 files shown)


HarelM:
Thanks for taking the time to open this PR! but I don't think this is a good UX.
I'll explain why: 
Maputnik is about allowing editors to see how their map would look. 
If I need to change the language all I need to do is load a different style.
If I want to edit a map related to a different language I usually won't need the UI of Maputnik to change as well, on the contrary, this would probably create worse experience as a designer as I'll have hard time finding the right buttons and panels...
I'm not saying we shouldn't think about better experience to support multiple map languages, but this is not exactly it.

Consider the following somewhat equivalent: you are writing a Google docs document, and every time you want to write in a different language all the UI changes according to the current language.

louwers:
@HarelM This does not affect any styles the user may load. It only affects which style is loaded when selecting the "ProtoMaps Light" default style.

louwers:
Updated title to be more clear.

src/config/styles.json
```diff
@@ -80,7 +80,7 @@
   {
     "id": "protomaps-light",
     "title": "Protomaps Light",
-    "url": "https://api.protomaps.com/styles/v2/light.json?key=d828297496b11844",
+    "url": "https://api.protomaps.com/styles/v4/light/$lang.json?key=d828297496b11844",
```
HarelM, line 83: I think the common usage is `{lang}` and not `$lang` in maplibre - glyphs, tiles etc use "{}" for parameters.

src/components/ModalOpen.tsx
```diff
@@ -56,6 +56,10 @@ type ModalOpenState = {
   activeRequestUrl?: string | null
 };
 
+function transformStyleUrl(styleUrl: string, {language}: { language: string }) {
+  return styleUrl.replace("$lang", language);
```
HarelM, line 60: What will happen if the url doesn't support the relevant language?

cypress/e2e/i18n.cy.ts
```diff
@@ -24,12 +26,42 @@ describe("i18n", () => {
...
+      when.select(selectorId, "ja");
+      then(get.elementByTestId(selectorId)).shouldHaveValue("ja");
 
       then(get.elementByTestId("nav:settings")).shouldHaveText("スタイル設定");
     });
   });
+
+  describe("load different style depending on selected language", () => {
+    const testLang = (lang: string) => {
+      cy.intercept('GET', `**/${lang}.json*`).as(`${lang}JsonRequest`);
```
HarelM, line 39: cy specific code should be moved to the driver.

HarelM:
Thanks for the info.
It's still not clear to me how one can use different languages from the protomaps styles, but it might be a different feature and out of scope for this PR.
I better understand the concept around this feature now, thanks!

PR #938, opened 2024-09-19
Add PMTiles support

pka:
Add support for pmtiles sources. Solves #807 

<strike>
There is still an error, probably when adding the source to the Maputnik sources:

```
Failed to process sources for 'pmtiles://https://example.com/data/switzerland.pmtiles' TypeError: NetworkError when attempting to fetch resource.
    fetchSources App.tsx:642
```

@bdon How did you solve this for https://editor.protomaps.com/ ?
</strike>

PR #962, opened 2024-12-26
Rebase #938 (pmtiles) onto main

### LINKED ISSUE #807: Support PMTiles in Editor
PMTiles from [Protomaps](https://protomaps.com/) are increasingly popular and a lot of other services are [supporting](https://github.com/felt/tippecanoe/releases/tag/2.17.0) it now. It has some major advantages in comparison to MBTiles. Styling the PMTiles are, however not possible yet, using the data set directly. 

MapLibre GL JS [supports PMTiles](https://protomaps.com/docs/frontends/maplibre). Therefore a switch to Maplibre GL JS in addition with the support of pmtiles URLs link ([like this](https://protomaps.github.io/PMTiles/?url=https%3A%2F%2Fprotomaps.github.io%2FPMTiles%2Fprotomaps%28vector%29ODbL_firenze.pmtiles#map=12.56/43.7798/11.24148))  would be super cool, so we can use PMTiles in Maputnik!

WebFreak001:
supersedes #938, close #938, close #807

also fixes the bug mentioned by @bdon (see 2nd commit) 
> There is a UI issue right now where the `pmtiles://` prefix sometimes appears in the modal data sources editor, sometime does not. Ideally the `pmtiles://` prefix is totally hidden from the editor, letting you naturally input a HTTP or HTTPS URL of a `.pmtiles` archive, and the internal Style JSON state has the full `pmtiles://https://...` string. This requires some changes to the existing patches, PR welcome.

cc @nyurik

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -43,7 +43,7 @@
         "maputnik-design": "github:maputnik/design#172b06c",
         "ol": "^6.14.1",
         "ol-mapbox-style": "^7.1.1",
-        "pmtiles": "^3.2.0",
+        "pmtiles": "^4.1.0",
         "prop-types": "^15.8.1",
         "react": "^18.2.0",
         "react-accessible-accordion": "^5.0.0",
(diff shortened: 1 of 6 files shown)


nyurik (approved):
this looks good to me. I'll let others to chime in, and merge a bit later unless something serious is discovered

HarelM:
This looks good, it works be better to have a test to cover the opening of this screen/flow.

bdon:
What I was referring to above: CleanShot 2024-12-27 at 13 58 07@2x

Where once a source is created, the `pmtiles://` prefix shows up. Though it may make more sense this way since it maps directly to how a developer would specify it in the Style JSON. So not an essential change.

WebFreak001:
I think it's good with `pmtiles://`, since this is also what you'll need to have in the final actual json - I just tested this with actually making a style and it worked very well and the protocol prefix helped me not forget it :+1:

WebFreak001:
rebase & improvements done in #938 now

PR #963, opened 2025-01-04
Add globe/mercator button

wipfli:
## Launch Checklist

Adds a button for choosing the projection globe or mercator. Appears under the light settings. Live demo is here: https://wipfli.github.io/maputnik/?layer=1455082206%7E0&modal=settings#1.02/0/0

image

 - [ ] Link to related issues.


.github/workflows/deploy.yml
```diff
@@ -2,7 +2,7 @@ name: deploy
 
 on:
   push:
-    branches: [ main ]
+    branches: [ globe ]
```
HarelM, line 5: This needs to be reverted?

HarelM:
I would recommend adding a test...

.github/workflows/deploy.yml
wipfli: Thanks yes reverted

wipfli:
Do you mind helping me with the test? I don't know cypress but happy to learn about it if you can help me a bit

HarelM:
sure, this repo has a good practice when it comes to end-to-end tests.
There's the driver which is the low level usage of cypress, usually, you won't need to touch it:
https://github.com/maplibre/maputnik/blob/main/cypress/e2e/maputnik-driver.ts
There's the helper which obscures how the UI looks and facilitate function to use in the tests so that the tests should look like a flow given-when-then style, but it's very slim in this case:
https://github.com/maplibre/maputnik/blob/main/cypress/e2e/maputnik-cypress-helper.ts
And there's the relevant test, since you added your functionality in a modal I would start by looking at a similar test here:
https://github.com/maplibre/maputnik/blob/main/cypress/e2e/modals.cy.ts

Let me know if the above is enough or not.
Thanks!

birkskyum:
@wipfli , could you add a third "Vertical Perspective" option? It's the "non-adaptive globe" projection that doesn't transition to mercator.

HarelM:
The maplibre packages' versions were updated to use latest ones, can you update the branch? It should reduce changes as part of this PR as well.

birkskyum:
Need rebase, since main is on GL JS 5

src/components/ModalSettings.tsx
```diff
@@ -208,6 +228,16 @@ class ModalSettingsInternal extends React.Component<ModalSettingsInternalProps>
           onChange={this.changeLightProperty.bind(this, "anchor")}
         />
 
+        <FieldEnum
+          label={"projection type"}
+          fieldSpec={latest.projection.type}
+          name="projection type"
+          value={projection.type as string}
+          options={Object.keys(latest.projection.type.values)}
```
birkskyum, line 236: options={['mercator', 'globe', 'vertical-perspective']}

HarelM:
Any updates on this?

wipfli:
I guess I should write the tests and then we could move on with this. I still want to help maintaining maputnik.

wipfli:
Thinking about this pull request again, I actually don't find the motivation to work on it further. Mainly because I don't use Maputnik myself.
I am closing this for now. If anyone else wants to pick it up again, feel free to reopen.

HarelM:
Thanks for taking the time to open this PR.
There are a few more additions needed now with the recent changes in version 5 I guess, but it'll have to wait for someone with the relevant time to allocate to it.

PR #967, opened 2025-01-16
add fallback behavior for `showOpenFilePicker` and `showSaveFilePicker`

josxha:
## Launch Checklist

 - [x] Link to related issues.

## Description

`showOpenFilePicker` and `showSaveFilePicker` are undefined on Firefox. With this pr, Maputnik uses the old behavior as a fallback. It keeps the naming "open" and "save" instead of "upload" and "download" to underline that the style stays within the browser and no actual upload happens.

@zstadler Could you give it a try, please?

## Related Issue

- fixes https://github.com/maplibre/maputnik/issues/966

## Visual Changes

The "Save as" button gets hidden if `showSaveFilePicker` is not available since it would have no use.

<table>
<tr>
<td>
Chrome
</td>
<td>
Firefox
</td>
</tr>
<tr>
<td>

</td>
<td>

</td>
</tr>
</table>

### DIFF
--- a/src/components/ModalExport.tsx
+++ b/src/components/ModalExport.tsx
@@ -86,6 +86,15 @@ class ModalExportInternal extends React.Component<ModalExportInternalProps> {
   async saveStyle() {
     const tokenStyle = this.tokenizedStyle();
 
+    // it is not guaranteed that the File System Access API is available on all
+    // browsers. If the function is not available, a fallback behavior is used.
+    if (typeof window.showSaveFilePicker !== "function") {
+      const blob = new Blob([tokenStyle], {type: "application/json;charset=utf-8"});
+      const exportName = this.exportName();
+      saveAs(blob, exportName + ".json");
+      return;
+    }
+
     let fileHandle = this.props.fileHandle;
     if (fileHandle == null) {
       fileHandle = await this.createFileHandle();
@@ -179,22 +188,18 @@ class ModalExportInternal extends React.Component<ModalExportInternalProps> {
         </div>
 
         <div className="maputnik-modal-export-buttons">
-          <InputButton
-            onClick={this.saveStyle.bind(this)}
-          >
+          <InputButton onClick={this.saveStyle.bind(this)}>
             <MdSave />
             {t("Save")}
           </InputButton>
-          <InputButton
-            onClick={this.saveStyleAs.bind(this)}
-          >
-            <MdSave />
-            {t("Save as")}
-          </InputButton>
+          {typeof window.showSaveFilePicker === "function" && (
+            <InputButton onClick={this.saveStyleAs.bind(this)}>
+              <MdSave />
+              {t("Save as")}
+            </InputButton>
+          )}
 
-          <InputButton
-            onClick={this.createHtml.bind(this)}
-          >
+          <InputButton onClick={this.createHtml.bind(this)}>
             <MdMap />
             {t("Create HTML")}
           </InputButton>
--- a/src/components/ModalExport.tsx
+++ b/src/components/ModalExport.tsx
@@ -123,10 +123,6 @@ class ModalExportInternal extends React.Component<ModalExportInternalProps> {
   }
 
   async createFileHandle(): Promise<FileSystemFileHandle | null> {
-    assert(
-      showSaveFilePickerAvailable,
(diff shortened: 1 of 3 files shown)


src/components/ModalExport.tsx
```diff
@@ -86,6 +86,15 @@ class ModalExportInternal extends React.Component<ModalExportInternalProps> {
   async saveStyle() {
     const tokenStyle = this.tokenizedStyle();
 
+    // it is not guaranteed that the File System Access API is available on all
+    // browsers. If the function is not available, a fallback behavior is used.
+    if (typeof window.showSaveFilePicker !== "function") {
```
HarelM, line 91: Store this in a variable to be used layer for readability?

zstadler:
LGTM!

src/components/ModalExport.tsx
josxha: I created an variable `showSaveFilePickerAvailable` for it. 👍

src/components/ModalExport.tsx
```diff
@@ -112,12 +122,16 @@ class ModalExportInternal extends React.Component<ModalExportInternalProps> {
     this.props.onOpenToggle();
   }
 
-  async createFileHandle() : Promise<FileSystemFileHandle | null> {
+  async createFileHandle(): Promise<FileSystemFileHandle | null> {
+    assert(
```
HarelM, line 126: Is this assert really needed?

src/components/ModalOpen.tsx
```diff
@@ -217,10 +244,16 @@ class ModalOpenInternal extends React.Component<ModalOpenInternalProps, ModalOpe
             <h1>{t("Open local Style")}</h1>
             <p>{t("Open a local JSON style from your computer.")}</p>
             <div>
-              <InputButton
-                className="maputnik-big-button"
-                onClick={this.onOpenFile}><MdFileUpload/> {t("Open Style")}
-              </InputButton>
+              {typeof window.showOpenFilePicker === "function" ? (
```
HarelM, line 247: Can we use if-else here? I think it will be clearer...

src/components/ModalOpen.tsx
```diff
@@ -168,6 +169,32 @@ class ModalOpenInternal extends React.Component<ModalOpenInternalProps, ModalOpe
     return file;
   }
 
+  // it is not guaranteed that the File System Access API is available on all
+  // browsers. If the function is not available, a fallback behavior is used.
+  onUpload = async (_: any, files: Result[]) => {
```
HarelM, line 174: [optional] Consider rnaming to `onFileChanged` to align with the event name.

josxha: done

src/components/ModalExport.tsx
josxha: nope, let's remove it.

src/components/ModalOpen.tsx
josxha: Is this the syntax you want to go with?

```jsx
  <div>
      {(() => {
        if (typeof window.showOpenFilePicker === "function") {
          return (
            <InputButton
                  className="maputnik-big-button"
                  onClick={this.onOpenFile}><MdFileUpload/> {t("Open Style")}
                </InputButton>
          )
        } else {
          return (
            <FileReaderInput onChange={this.onFileChanged} tabIndex={-1} aria-label={t("Open Style")}>
                  <InputButton className="maputnik-upload-button"><MdFileUpload /> {t("Open Style")}</InputButton>
                </FileReaderInput>
          )
        }
      })()}
    </div>
```

HarelM: Is the lambda mandatory? I recall that react has a simpler way to add if inside the render method...?

josxha: Not sure how? I can't find anything about if in the [react docs](https://legacy.reactjs.org/docs/conditional-rendering.html) nor [stackoverflow](https://stackoverflow.com/questions/40477245/is-it-possible-to-use-if-else-statement-in-react-render-function). Using a tenary operator is the shortest way. Alternatively, we could move the html block before the return statement and use it from a variable.

HarelM (approved):

PR #1015, opened 2025-01-21
Use new scss syntax

birkskyum:
SCSS is changing, giving a lot of runtime console warnings.

https://sass-lang.com/documentation/breaking-changes/import/

So we have to use `@use` instead of `@import` and a few other things. It's aligning more with css

Their migration tool did almost everything, and console warnings described remaining Changes
- https://sass-lang.com/documentation/breaking-changes/import/#automatic-migration

### DIFF
--- a/src/components/SmallError.scss
+++ b/src/components/SmallError.scss
@@ -1,7 +1,7 @@
-@import '../styles/vars';
+@use '../styles/vars';
 
 .SmallError {
   color: #E57373;
-  font-size: $font-size-6;
-  margin-top: $margin-2
+  font-size: vars.$font-size-6;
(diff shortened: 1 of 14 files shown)


HarelM (commented):
Oh man, so much css, this really needs to be replaced by a UI component library...
Make sure this looks as expected as e2e tests do not cover this kind of behavior change.

birkskyum:
I've made visual comparisons, and they check out unchanged. It seems to be just a new import syntax where the style vars are scoped to modules.

Screenshot 2025-01-21 at 22 18 48

Screenshot 2025-01-21 at 22 18 55

HarelM (approved):

PR #1018, opened 2025-01-21
Use modern jsonlint fork

birkskyum:
We have a few vulnerabilities mentioned when installing, due to obsolete versions of Underscore, which comes from a 10y old version of jsonlint.

This PR swap the old jsonlint from 2015:
https://github.com/josdejong/jsonlint

With this one:
https://github.com/prantlf/jsonlint

## Launch Checklist

 - [ ] Link to related issues.


HarelM:
This looks good assuming it is working as expected, were there other mentioning of this package that can be removed? Maybe in the rollup configuration?
Can this be done to the style-spec package as well?

src/libs/codemirror-mgl.ts
```diff
@@ -41,19 +28,10 @@ CodeMirror.registerHelper("lint", "json", (text: string) => {
...
-    const loc = hash.loc;
-    found.push({
-      from: CodeMirror.Pos(loc.first_line - 1, loc.first_column),
-      to: CodeMirror.Pos(loc.last_line - 1, loc.last_column),
-      message: str
-    });
-  };
   try {
-    parser.parse(text);
+    parse(text);
```
HarelM, line 34: Isn't the error handing logic is missing? 
Probably worth adding a test to make sure this is working as expected before and after this change.

birkskyum:
I got the error reporting workign.

So it's using the error from the catch(err)  now instead of the parseError hack

Screenshot 2025-01-21 at 22 37 07

birkskyum:
I abstracted this error handling out. It's the exact same function we had twice

birkskyum:
@HarelM , I managed to add a test that brakes the json, and checks that the red error circle appear in the side of the editor

cypress/downloads/downloads.html
HarelM, line 1: What's this file? Is it a binary or text file?

cypress/e2e/layers.cy.ts
```diff
@@ -494,4 +494,33 @@ describe("layers", () => {
...
+          },
+        ],
+      });
+
+      const sourceText = get.elementByText('"source"');
+      
+      sourceText.click();
+      sourceText.type("\"");
+      
+      const error = cy.get('.CodeMirror-lint-marker-error');
```
HarelM, line 522: "cy." Code should be in the helper/driver.

HarelM:
Thanks!

cypress/downloads/downloads.html
birkskyum: don't know where it came from. something cypress made. removed it.

cypress/e2e/layers.cy.ts
birkskyum: used get.element instead

cypress/e2e/layers.cy.ts
```diff
@@ -494,4 +494,33 @@ describe("layers", () => {
       ).shouldBeVisible();
     });
   });
+
+
+  describe.only("layereditor jsonlint should error", ()=>{
```
HarelM, line 499: remove only...

HarelM (approved):
Just remove the "only" and you can merge it, thanks!!

cypress/e2e/layers.cy.ts
birkskyum: good catch

PR #1021, opened 2025-01-21
Fix insertion of access tokens, when swapping renderer

birkskyum:
Going from e.g. MapTiler to OpenLayers and back will lose the maptlier key.

This code finds the urls in the style that has "{key}" and insert the correct API keys

Fixes the error reported here, cc @nyurik 
- Fixes https://github.com/maplibre/maputnik/issues/874#issuecomment-2605896666

Related to:
- https://github.com/maplibre/maputnik/issues/869

## Launch Checklist

 - [ ] Link to related issues.

### DIFF
--- a/cypress/e2e/modals.cy.ts
+++ b/cypress/e2e/modals.cy.ts
@@ -1,4 +1,5 @@
 import { MaputnikDriver } from "./maputnik-driver";
+import tokens from "../../src/config/tokens.json" with {type: "json"};
 
 describe("modals", () => {
   const { beforeAndAfter, when, get, then } = new MaputnikDriver();
@@ -238,7 +239,7 @@ describe("modals", () => {
 
 
 
-    it.only("style renderer change", () => {
+    it("inlcude API key when change renderer", () => {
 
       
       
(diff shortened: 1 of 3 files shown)


src/components/App.tsx
```diff
@@ -781,6 +803,12 @@ export default class App extends React.Component<any, AppState> {
   getInitialStateFromUrl = (mapStyle: StyleSpecification) => {
     const url = new URL(location.href);
     const modalParam = url.searchParams.get("modal");
+
+
+    console.log("HERE12", url)
```
HarelM, line 808: This shouldn't be here...

src/components/App.tsx
```diff
@@ -362,6 +362,25 @@ export default class App extends React.Component<any, AppState> {
         [property]: value
       }
     }
+
+
+    // For the style object, find the urls that has "{key}" and insert the correct API keys
```
HarelM, line 367: Why is this call not part of the `onStyleChanged` method?

HarelM:
Is there a test we can add to make sure this behavior is kept in the future?

birkskyum:
I'm not sure how to test this through dom/cypress. It's about network calls and visual output on canvas.

HarelM:
You can use Cypress to intercept network calls I believe.
Also you can check what is stored in the lical storage, or is the local storage content "correct"?

birkskyum:
@HarelM , figured out the test now. It checks the requests made after the render change has the access token instead of just `{key}`

cypress/e2e/modals.cy.ts
```diff
@@ -217,6 +218,52 @@ describe("modals", () => {
...
+
+    it("inlcude API key when change renderer", () => {
+
+      
+      
+      when.click("modal:settings.close-modal")
+      when.click("nav:open");
+
+
+      // when.click("aria-label="MapTiler Basic"
```
HarelM, line 232: Remove comments

cypress/e2e/modals.cy.ts
```diff
@@ -217,6 +218,52 @@ describe("modals", () => {
...
+
+      cy.get('[aria-label="MapTiler Basic"]').should('exist').click();
```
HarelM, line 234: Use helper or driver

HarelM:
The test needs some cleaning up, otherwise this looks good, thanks!

cypress/e2e/modals.cy.ts
birkskyum: Just so I don't put things completely random, what's the gist of this driver / helper? I assume it's to increase code reuse, and the helper is in the driver instance? modal/maputnik driver, what's the difference there?

HarelM: Check out the architecture documents in the readme of the following package:
https://www.npmjs.com/package/@shellygo/cypress-test-utils
The driver should communicate with cypress - have `cy.something` methods (in most cases, there's no need to add anything here as part of this project)
The helper should facilitate "PM" like options - clickThat, openModal, navigateToAScreen, and the test itself should be using the helper to create a test that sounds like an acceptance criteria - `given.somthing(); when.IDoSomething(); then.IshouldBeSomething();`
It might not be perfect or exactly that in all tests, but this is the concept.

birkskyum: Getting a hang of it - I managed to do that now.

HarelM:
Firefox and cypress do not play together well, this is not the fire time the tests are passing with chrome but not with firefox, maybe we should drop support for firefox and hope it works well, IDK...

birkskyum:
@HarelM , I've removed firefox in this pr too.

HarelM:
Great! Thanks!

birkskyum:
Not sure why it's still showing an "E2E tests using firefox"

birkskyum:
@HarelM , is this PR approved?

HarelM:
It was required in the setting of the branch protection, I've removed it and fixed it now.

HarelM:
@birkskyum can you take a look at the added cypress test here?
I think it causes some CI instability and prevents automatic dependencies updates...

PR #1025, opened 2025-01-23
Add support for local pmtiles tile package (via drag and drop)

prusswan:
This adds support for using local pmtiles tile packages that can be loaded via drag and drop (and possibly file picker). The behavior is very similar to https://pmtiles.io. These tile packages can already be referenced/used as a data source in the style, e.g. Tile URL: `pmtiles://filename.pmtiles/{z}/{x}/{y}`, just lacking a way to load the content.

**Visual changes:**

Added new spot in the top toolbar for dropping pmtiles package, also works with Inspect:
image

## Launch Checklist

 - [ ] Link to related issues.


src/components/AppToolbar.tsx
```diff
@@ -289,6 +292,15 @@ class AppToolbarInternal extends React.Component<AppToolbarInternalProps> {
             <MdHelpOutline />
             <IconText>{t("Help")}</IconText>
           </ToolbarLink>
+
+          <Dropzone onDrop={this.props.onFileSelected}>
+            {({getRootProps, getInputProps}) => (
+              <div {...getRootProps({className: 'dropzone maputnik-toolbar-link'})}>
+                <input {...getInputProps()} />
+                Drop file here
```
HarelM, line 300: This needs translation...

src/components/MapMaplibreGl.tsx
```diff
@@ -134,7 +137,21 @@ class MapMaplibreGlInternal extends React.Component<MapMaplibreGlInternalProps,
...
+    if (this.props.file) {
+      const file = this.props.file;
+      this.state.protocol!.add(file); // this is necessary for non-HTTP sources
+
+      if (map) {
+        file.getMetadata().then((metadata: any) => {
+          const layerNames = metadata.vector_layers.map((e: LayerSpecification) => e.id);
+
+          map.style.sourceCaches["source"]._source.vectorLayerIds = layerNames;
+          console.log("layerNames for inspect:", layerNames);
```
HarelM, line 150: Remove console. Loh

src/components/App.tsx
```diff
@@ -131,6 +131,7 @@ type AppState = {
     debug: boolean
   }
   fileHandle: FileSystemFileHandle | null
+  file: PMTiles | null
```
HarelM, line 134: This doesn't look like a good name for the relevant state...

src/components/AppToolbar.tsx
prusswan: This came from https://react-dropzone.js.org/#section-class-components, or do you mean i18n?

src/components/App.tsx
prusswan: how about localPMTiles?

src/components/AppToolbar.tsx
```diff
@@ -17,6 +17,8 @@ import maputnikLogo from 'maputnik-design/logos/logo-color.svg?inline'
 import { withTranslation, WithTranslation } from 'react-i18next';
 import { supportedLanguages } from '../i18n';
 
+import Dropzone from 'react-dropzone'; // for class components
```
josxha, line 20:
```suggestion
import Dropzone from 'react-dropzone';
```
Is this comment meaningful?

src/components/MapMaplibreGl.tsx
```diff
@@ -74,6 +75,7 @@ type MapMaplibreGlState = {
   geocoder: MaplibreGeocoder | null;
   zoomControl: ZoomControl | null;
   zoom?: number;
+  protocol: Protocol | null;
```
josxha, line 78: A name that makes it more clear that this is a pmtiles Protocol would be good.

src/components/AppToolbar.tsx
josxha: It not clear that this Dropzone relates only to a pmtiles file. As a user I'd expect that I could drag and drop a style json. 
If feasable, how about a full screen drop zone, where the user can drag and drop the file anywhere on the window?

prusswan: @josxha I have updated i18n to make it clear that this is only for PMTiles

src/components/AppToolbar.tsx
prusswan: This came from `react-dropzone` docs. Functional/class components require different imports (useDropzone/Dropzone respectively)

josxha: Yes, please.

src/components/AppToolbar.tsx
```diff
@@ -103,6 +105,7 @@ type AppToolbarInternalProps = {
   onSetMapState(mapState: MapState): unknown
   mapState?: MapState
   renderer?: string
+  onFileSelected(...args: unknown[]): unknown
```
HarelM, line 108: Don't we know the arguments type? This isn't great from a typescript perspective.

src/components/App.tsx
```diff
@@ -881,6 +884,14 @@ export default class App extends React.Component<any, AppState> {
     });
   }
 
+  onFileSelected = (e: File[]) => {
```
HarelM, line 887: Why os this here while the drop zone is not part of this component? I find it a bit strange...

prusswan: I will try to move this to AppToolbar, and pass setLocalPMTiles as a prop to AppToolbar (which handles onFileSelected then calls this.props.setlocalPMTiles). Is this what you have in mind?

src/components/AppToolbar.tsx
```diff
@@ -103,6 +105,7 @@ type AppToolbarInternalProps = {
   onSetMapState(mapState: MapState): unknown
   mapState?: MapState
   renderer?: string
+  setLocalPMTiles(...args: File[]): unknown
```
HarelM: I think `onLocalPMTilesSelected` would describe what this method is for.

src/components/App.tsx
```diff
@@ -881,6 +884,13 @@ export default class App extends React.Component<any, AppState> {
     });
   }
 
+  setLocalPMTiles = (file: File) => {
+    const pmt = new PMTiles(new FileSource(file));
```
HarelM, line 888: I'm not sure if this should be here on right after selecting the file and raising an event with the selected `PMTiles` class.
Either way this is fine. your call.

HarelM (approved):

HarelM:
Looking at this holistically, I'm not sure this is the right way to achieve this.
Add sources as files may require support other file types like shp files, geojson etc, and the idea of this editor is to allow editing the style and I'm not sure it should handle loading of file like what is suggested here.
I don't know...

prusswan:
> Looking at this holistically, I'm not sure this is the right way to achieve this. Add sources as files may require support other file types like shp files, geojson etc, and the idea of this editor is to allow editing the style and I'm not sure it should handle loading of file like what is suggested here. I don't know...

I agree that this functionality is only feasible for PMTiles (up to a certain size) and is aided by the design and implementation of PMTiles library itself.

HarelM:
I've posted a message on slack, let's see what the community thinks about this feature.

louwers (commented):
In my opinion there should be some way this feature is documented. Ideally it should be made discoverable. I suggest that when you drag anything onto the screen, you see a message saying what file types are supported. If you drag something that has an unsupported file type, you should get a message which file types are supported.

We can then extend the supported file types, for example, also allowing a style JSON to be dragged-and-dropped.

louwers (changes requested):
Should be tested with Cypress as well.

HarelM:
I'm not sure It can be tested since dragging a file into the browser requires the ability to connect with the OS or something.
Not sure, but I won't be surprised if this can't be tested.

prusswan:
> In my opinion there should be some way this feature is documented. Ideally it should be made discoverable. I suggest that when you drag anything onto the screen, you see a message saying what file types are supported. If you drag something that has an unsupported file type, you should get a message which file types are supported.

I have set the file type to allow for .pmtiles only, this is reflected in the file picker. If invalid files are dragged instead, file rejection messages will be shown in console and alert dialog. 

As for documentation, this can be a new section under the wiki?

louwers:
@HarelM @prusswan https://github.com/cypress-io/cypress-example-recipes/tree/master/examples/testing-dom__drag-drop

louwers:
Sorry, should have been more specific. This one looks like it should work:

> Selects a file or files in an HTML5 input element or simulates dragging a file or files into the browser.

HarelM:
Ahh, found what you were referring to here in the Cypress docs:
https://docs.cypress.io/api/commands/selectfile

cy.document().selectFile('file.json', { action: 'drag-drop' })

prusswan:
e2e test added: I wanted to test for popup but I simply could not get mouseover to work, so I settled for image diff. Hope this is not too big of change

cypress.config.ts
```diff
@@ -1,5 +1,6 @@
 import { defineConfig } from "cypress";
 import { createRequire } from "module";
+import getCompareSnapshotsPlugin from 'cypress-image-diff-js/plugin';
```
HarelM, line 3: Is there a simpler solution without image diff? Checking that there's no error in case the file type is OK and check that the error appears when the file isn't?

src/locales/he/translation.json
```diff
@@ -35,6 +35,7 @@
   "View": "תצוגה",
   "Color accessibility": "נגישות צבעים",
   "Help": "עזרה",
+  "Drop PMTiles file here": "__STRING_NOT_TRANSLATED__",
```
HarelM, line 38:
```suggestion
  "Drop PMTiles file here": "גררו לפה קובץ PMTiles",
```

src/locales/he/translation.json
```diff
@@ -81,6 +82,7 @@
   "Close modal": "סגירת חלונית",
   "Debug": "דיבאג",
   "Options": "אפשרויות",
+  "<0>Open in OSM</0> &mdash; Opens the current view on openstreetmap.org": "__STRING_NOT_TRANSLATED__",
```
HarelM, line 85:
```suggestion
```

src/components/AppToolbar.tsx
```diff
@@ -134,6 +137,19 @@ class AppToolbarInternal extends React.Component<AppToolbarInternalProps> {
...
+    const file = e[0];
+    this.props.onLocalPMTilesSelected(file);
+  }
+
+  onFileRejected = (r: FileRejection[]) => {
+    const errorMessageLine = r.map(e => {
+      return e.errors.map(f => f.message).join("\n")
+    }).join("\n");
+    console.error("Dropzone file rejected:", errorMessageLine);
+    alert(errorMessageLine);
```
HarelM, line 150: I'm sure there are better ways to notify the user of a problem, not to mention this is not translated...

src/components/AppToolbar.tsx
```diff
@@ -134,6 +137,19 @@ class AppToolbarInternal extends React.Component<AppToolbarInternalProps> {
...
+  onFileSelected = (e: File[]) => {
+    const file = e[0];
+    this.props.onLocalPMTilesSelected(file);
+  }
+
+  onFileRejected = (r: FileRejection[]) => {
+    const errorMessageLine = r.map(e => {
+      return e.errors.map(f => f.message).join("\n")
+    }).join("\n");
+    console.error("Dropzone file rejected:", errorMessageLine);
```
HarelM, line 149: I don't see a reason to both print to the console and notify the user, so this can be removed I guess.

src/components/AppToolbar.tsx
```diff
@@ -134,6 +137,19 @@ class AppToolbarInternal extends React.Component<AppToolbarInternalProps> {
     }
   }
 
+  onFileSelected = (e: File[]) => {
+    const file = e[0];
+    this.props.onLocalPMTilesSelected(file);
+  }
+
+  onFileRejected = (r: FileRejection[]) => {
+    const errorMessageLine = r.map(e => {
```
HarelM, line 146: Is it critical to collect all the error messages?

src/components/MapMaplibreGl.tsx
```diff
@@ -134,7 +137,23 @@ class MapMaplibreGlInternal extends React.Component<MapMaplibreGlInternalProps,
...
       }, 500);
     }
-    
+
+    if (this.props.localPMTiles) {
+      const file = this.props.localPMTiles;
+      this.state.pmtilesProtocol!.add(file); // this is necessary for non-HTTP sources
+
+      if (map) {
+        file.getMetadata().then((metadata: any) => {
```
HarelM, line 146: Does metadata has a better type? `topojson` maybe?

src/components/AppToolbar.tsx
prusswan: I can just show the first error and apply translation for it, if that is okay with you.

HarelM: Are you planning on translating all the errors?
I would simply present a generic error message to the user about failure to open the provided file.

prusswan: I think I can log the errors from react-dropzone to console, and show a translatable error message to the user.

src/components/MapMaplibreGl.tsx
prusswan: Added an interface for Metadata (borrowed from pmtiles code)

PR #1079, opened 2025-02-24
Group dependabot updates

### LINKED ISSUE #1020: Make dependabot PRs contain multiple updates
This will reduce the number of PRs, consolidating many dependency changes into one. See similar issue at https://github.com/maplibre/martin/issues/1622

nyurik:
Ensure dependabot creates updates in batches

Fixes #1020

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,8 +9,16 @@ updates:
     directory: "/" # Location of package manifests
     schedule:
       interval: "daily"
-    open-pull-requests-limit: 20
     versioning-strategy: increase
+    groups:
+      all-actions-version-updates:
+        applies-to: version-updates
+        patterns:
+          - "*"
+      all-actions-security-updates:
+        applies-to: security-updates
+        patterns:
+          - "*"
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:


HarelM:
Are the batches related to a npm scope? How are these grouped? 
In general, besides the fact that CI is currently unstable, the dependabot PRs are merged automatically and the fact that they are not grouped allows to know which package broke the build, so it's not a bad thing...

nyurik:
the batch is always per updater, i.e. NPM has one group, whereas if this project had Java, that would be a separate group.  The groups are created daily. You do have a good point about breakage detection, but it is only useful if they break all the time.  Usually, most batches will pass. In the rare case when they do not, we can fairly easily tell which component broke, and update them by hand. Overall, I think the net-gain will be significantly higher because it will lessen the mental load and flood of PRs

louwers (approved):

birkskyum:
With the `    open-pull-requests-limit: 20` removed, what's the default value it fall back on?

nyurik:
i am begining to question this PR myself tbh... having come from the Rust background, when dependencies are updated daily and there is a very high expected stability, in js it seems everything is far more.... unpredictable... so perhaps it is too soon to add this... I will just have to filter my mailbox from all the junk

HarelM:
Closing, as I believe this is not the right approach to javascript packages...

PR #1080, opened 2025-02-24
Add precommit check

nyurik:
Keeps the repo clean, same as several other of our repos


HarelM:
Can we use a CI check instead of pre-commit?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,21 @@
+# See https://pre-commit.com for more information
+# See https://pre-commit.com/hooks.html for more hooks
+
+repos:
```
louwers, line 4: Might want to set the autoupdate frequency to quarterly.

https://pre-commit.ci/#configuration-autoupdate_schedule

nyurik:
Per our online chat - pre-commit is the name of the CI, not the `.git/hooks/before-push` hook.  So yes, it is CI, just provided by a different provider than github.

.pre-commit-config.yaml
nyurik: i set it to monthly, rare enough i think

nyurik: just curious - why quarterly?  the default weekly seems "good enough" - as it probably won't be updated that often anyway

birkskyum:
This is all I get...
Screenshot 2025-02-25 at 10 54 34

birkskyum:
@nyurik , what's the status of this? Would you prefer we merge this config, so it's all going, or that we disable the github app until this PR is ready?

nyurik:
@birkskyum try it now

PR #1081, opened 2025-02-25
unflake api key test

birkskyum:
This adds a small 1s timeout after changing to the maptiler style.

The test is flaky because it try to move on before the style is loaded


nyurik (approved):
lgtm

PR #1102, opened 2025-03-14
fix url to to versatiles colorful

yetzt:
the stylejson url to versatiles colorful has changed from `https://tiles.versatiles.org/assets/styles/colorful.json` to `https://tiles.versatiles.org/assets/styles/colorful/style.json`


yetzt:
thank you :)

PR #1114, opened 2025-03-26
fix: fix filter-editor style

BlackGanglion:
Include before/after visuals or gifs if this PR includes visual changes.

### Before
image

### After
image

### DIFF
--- a/src/styles/_filtereditor.scss
+++ b/src/styles/_filtereditor.scss
@@ -16,11 +16,10 @@
 
 .maputnik-filter-editor-single {
   display: flex;
+  column-gap: 6px;
 }
 
 .maputnik-filter-editor-operator {
-  margin: 0 6px;
-
   .maputnik-select {
     width: 100%;
   }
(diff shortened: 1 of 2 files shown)


src/styles/_filtereditor.scss
```diff
@@ -67,17 +60,13 @@
...
-  margin-right: 1.5%;
-}
+.maputnik-filter-editor-block {
+  display: flex;
+  margin-bottom: 2px;
 
-.maputnik-filter-editor-block-content {
-  display: inline-block;
-  width: 92.5%;
+  &-action {
```
HarelM, line 67: What will happen for RTL content for this rule?

HarelM:
I'm thinking if adding support for tailwindCSS as part of this PR makes sense?

src/styles/_filtereditor.scss
BlackGanglion: I reconsidered RTL content

image
image

BlackGanglion:
> I'm thinking if adding support for tailwindCSS as part of this PR makes sense?

Supporting tailwindCSS is a big project. I suggest fixing the existing issues first.

HarelM (approved):

PR #1253, opened 2025-07-03
Handle QuotaExceededError in StyleStore

louwers:
## Launch Checklist

When localStorage is full you start getting a "QuotaExceededError". RIght now Maputnik does not handle this situation gracefully, it just fails loading the style with a non-descriptive error message.

This PR purges localStorage and tries again when this particular error happens.

It still does not show a descriptive error message. If you try to load a style that is larger than what localStorage can handle, it will still fail with a non-descriptive error message.

Increased the size of `example-style.json` so that it causes a QuotaExceededError when running the regression test (try it before and after this PR).

 - [x] Link to related issues. N/A
 - [x] Include before/after visuals or gifs if this PR includes visual changes. N/A


nyurik (approved):

HarelM:
I'm guessing indexDB would be a better long term solution, wouldn't it?

louwers:
Probably. But honestly I don't even know what the StyleStore is used for?

HarelM:
For keeping the work of a user between refreshes of the browser.

PR #1263, opened 2025-07-04
Replace react-autosuggest

louwers:
## Summary
- switch InputAutocomplete to `downshift`
- remove obsolete `react-autocomplete` dependency
- document autocomplete change in the changelog

## Testing
- `npm run lint`
- `npm run build`

------
https://chatgpt.com/codex/tasks/task_e_68685a78c2d483319c068f813723c1a7


HarelM:
This is possible the most important replacement needed in order to remove the console log. 
Is it possible to take an auto complete from a UI component library so that we can continue adopting other components? 
Also a test would be great here. 

Looks good! Thanks for pushing this forward!

louwers:
@HarelM I think choosing a component library is a bigger task than just replacing a deprecated autocomplete library.

I think there is something to be said for sticking to using custom components. It is hard to have a consistent look-and-feel with a mix and match. Also I think using a component library is not necessarily less maintenance work, because they contain the kitchen sink in terms of functionality and these libraries tend to come and go as well.

In any case, I set up component testing and added a test for `InputAutocomplete`.

src/components/InputAutocomplete.cy.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { mount } from 'cypress/react'
+
+const fruits = ['apple', 'banana', 'cherry'];
+
+describe('<InputAutocomplete />', () => {
+  it('filters options when typing', () => {
+    mount(
+      <InputAutocomplete aria-label="Fruit" options={fruits.map(f => [f, f])} />
+    );
+    cy.get('input').focus();
```
HarelM, line 11: Other tests have the driver and helper pattern, shouldn't we use this pattern here as well? 
I'm ok either way, but thought I should bring this up.

louwers: Do you mean use `cy.mount`? I could not get TypeScript to shut up.

HarelM: No, checkout how the e2e are written, they use some abstractions, which can be used in component tests as well I belive.

PR #1264, opened 2025-07-05
Add AGENTS.md

louwers:
This file is read by LLM agents when working on the codebase: https://agentsmd.net/

With this file you don't need to copy paste these instructions in every prompt.


birkskyum:
I reckon we can always elaborate on variations for other OS's

louwers:
@birkskyum I think all these LLM agents use Linux...

PR #1268, opened 2025-07-05
Fix duplicate layer rename error

louwers:
## Summary
image

- warn when changing layer ID to duplicate
- test duplicate layer rename error
- document the change in the changelog

## Testing
- `npm run lint`
- `npm run build`
- `xvfb-run -a npm run test` *(fails: Cypress couldn't verify that server is running)*

------
https://chatgpt.com/codex/tasks/task_e_686907f11134833184ffde804d8b364d

### DIFF
--- a/src/components/LayerEditor.tsx
+++ b/src/components/LayerEditor.tsx
@@ -135,6 +135,7 @@ class LayerEditorInternal extends React.Component<LayerEditorInternalProps, Laye
     if(this.props.layer.metadata) {
       comment = (this.props.layer.metadata as any)['maputnik:comment']
     }
+    const t = this.props.t;
     const {errors, layerIndex} = this.props;
 
     const errorData: {[key in LayerSpecification as string]: {message: string}} = {};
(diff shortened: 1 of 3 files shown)


src/components/LayerEditor.tsx
```diff
@@ -157,53 +158,57 @@ class LayerEditorInternal extends React.Component<LayerEditorInternalProps, Laye
...
-      {this.props.layer.type !== 'background' && <FieldSource
-        error={errorData.source}
-        sourceIds={Object.keys(this.props.sources!)}
-        value={this.props.layer.source}
-        onChange={v => this.changeProperty(null, 'source', v)}
-      />
-      }
-      {['background', 'raster', 'hillshade', 'heatmap'].indexOf(this.props.layer.type) < 0 &&
+    case 'layer': {
+      const duplicateId = errorData.id && /duplicate layer id/.test(errorData.id.message);
```
HarelM, line 162: Is that the best way to do this? Seems odd...

louwers: I agree. It does not seem to be able to elegantly solve the problem.

PR #1350, opened 2025-09-08
Use desktop build inside docker

HarelM:
## Launch Checklist

This PR uses the maputnik CLI inside the docker container to allow loading of files into the container and watch for changes.


HarelM:
I don't see a reason to create two docker images.
If you run the docker image without any parameters you get the exact same behavior as with the nginx (previous docker image) as the go mainly wraps the built html/css/js files and serve them.

The advantage of this PR is that you can now also mount styles and be able to edit them from the browser (although this is somewhat supported with the file API that was added to the browser).
You can also serve some other static files (pmtiles, glyphs, sprites, tiles) with this image.

The bottom line is that if we support the "desktop" version, I think it makes more sense to have that inside the docker image and not only the html/css/javascript.

louwers:
> The bottom line is that if we support the "desktop" version

Do we want to support it? It seems to depend on Golang dependencies such as [GeertJohan/go.rice](https://github.com/GeertJohan/go.rice) and [gorilla/handlers](github.com/gorilla/handlers) that do not appear to be actively maintained.

Maputnik is inherently a web app which has many advantages such as ease of deployment, that we lose when we try to shoehorn it into a desktop app. And I think the web platform is powerful enough that we can deliver the same functionality.

Personally I would see more future in integrating the File System Access API you alluded to.

HarelM:
I generally agree that I don't see a lot of value in the desktop app, and yet, this was one of the only contribution we got to maputnik, so I believe there are people who are interested in using it.
I don't mind removing all the desktop stuff, which I basically did when migrating maputnik to maplibre org, but it was added back, so there's probably a merit to it.

From a local development point of view, I don't see a real problem as you basically develop with vite, the CI takes care of the rest, and since it's pretty stable and doesn't require changes I don't see a lot of development overhead.

HarelM:
Besides obviously right now, that I'm trying to understand the boot sequence of maputnik, but that's not all related to the desktop...

louwers:
I think [long term](https://developer.mozilla.org/en-US/docs/Web/API/FileSystemObserver) there may be a better approach to reading/watching files from disk than a desktop app. That's why I suggested keeping the Dockerfile separate, because I think at some point we want to get rid of the Golang based desktop app.

FWIW here are some numbers:

2.1.1

desktop.zip (34.08 MiB) - downloaded 1,529 times. Last updated on 2024-08-29

1.7.0

maputnik-darwin.zip (8.26 MiB) - downloaded 1,298 times. Last updated on 2020-07-24
maputnik-linux.zip (8.31 MiB) - downloaded 2,267 times. Last updated on 2020-07-24
maputnik-windows.zip (8.08 MiB) - downloaded 3,226 times. Last updated on 2020-07-24

maplibre.org/maputnik

July -  251 visits
August - 263 visits

<details><summary>CloudFlare Query</summary>

```graphql
query {
  viewer {
    zones(filter: {zoneTag: "01da56efa6ceb0e5d428095125798b1b"}) {
      httpRequestsAdaptiveGroups(
        filter:{
          date_gt: "2025-08-15"
          date_lt: "2025-08-31"
          clientRequestPath_like: "/maputnik/%"
          requestSource: "eyeball"
        }
        limit: 100
      ) {
        count
      }
    }
  }
}
```
<details>

HarelM:
I think that if and when this will happen we will decide how to change the docker file along other files.
Keeping two images doesn't make a lot of sense maintenance-wise (although it a file of like 20 lines I don't see a value in keeping 2 in this repo).
Having this docker file also allow building the desktop app without installing go, which is a good added value I believe.

PR #1351, opened 2025-09-08
Move style and store initialization to mount method

HarelM:
This is in order to reduce warnings in the console for React 19 usage.
This removes the deprecated defaultProp and also move all the store initialization logic out of the App.tsx file, keeping it a lot more clean.
It removes the `debug` flag from the supported urls along with the `localport` and `localhost`, which I'm not sure if and how they were ever used.
The tests are using the `style` url, so I think it is covered in terms of tests.
It also improves some typings along the project.
It removes some callbacks from the code and moves to use promises.

## Launch Checklist

Before:
image

After:
image

### DIFF
--- a/cypress/e2e/maputnik-driver.ts
+++ b/cypress/e2e/maputnik-driver.ts
@@ -126,7 +126,7 @@ export class MaputnikDriver {
         break;
       }
       if (zoom) {
-        url.hash = `#${zoom}/41.3805/2.1635`;
+        url.hash = `${zoom}/41.3805/2.1635`;
       }
       this.helper.when.visit(url.toString());
       if (styleProperties) {


src/libs/store/style-store-factory.ts
```diff
@@ -0,0 +1,30 @@
+import { IStyleStore, OnStyleChangedCallback } from "../definitions";
+import { getStyleUrlAndRemoveItIfNeeded, loadStyleUrl } from "../urlopen";
+import { ApiStyleStore } from "./apistore";
+import { StyleStore } from "./stylestore";
+
+declare const __mode__: 'desktop' | 'production' | 'development';
```
louwers: Use `import.meta.env.MODE` instead? https://vite.dev/guide/env-and-mode.html#built-in-constants

src/libs/store/style-store-factory.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export async function createStyleStore(onStyleChanged: OnStyleChangedCallback): Promise<IStyleStore> {
```
louwers, line 7: Why does creating a style store load a style?

Does this have to be async?

src/libs/store/stylestore.ts
```diff
@@ -78,17 +74,21 @@ export class StyleStore {
   }
 
   // Find the last edited style
-  latestStyle(cb: (...args: any[]) => void) {
-    if(this.mapStyles.length === 0) return loadDefaultStyle(cb)
+  async getLatestStyle(): Promise<StyleSpecificationWithId> {
+    if(this.mapStyles.length === 0) {
```
louwers, line 78: Aside, but should we maybe add a formatter?

src/libs/urlopen.ts
```diff
@@ -1,30 +1,27 @@
 import style from './style'
+import { StyleSpecificationWithId } from './definitions';
 
-export function initialStyleUrl() {
+export function getStyleUrlAndRemoveItIfNeeded(): string | null {
```
louwers: Remove from where?

cypress/e2e/maputnik-driver.ts
```diff
@@ -110,19 +110,19 @@ export class MaputnikDriver {
       styleProperties: "geojson" | "raster" | "both" | "layer" | "",
       zoom?: number
     ) => {
-      let url = "?debug";
+      let url = "";
       switch (styleProperties) {
       case "geojson":
-        url += `&style=${baseUrl}geojson-style.json`;
+        url += `?style=${baseUrl}geojson-style.json`;
```
louwers: Probably makes sense to use [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams)...

src/libs/urlopen.ts
HarelM: From address bar.
I can add that to the method name.

src/libs/store/style-store-factory.ts
HarelM: Thanks!
Done in 7093ebf

src/libs/store/style-store-factory.ts
```diff
@@ -3,13 +3,11 @@ import { getStyleUrlAndRemoveItIfNeeded, loadStyleUrl } from "../urlopen";
...
 import { StyleStore } from "./stylestore";
 
-declare const __mode__: 'desktop' | 'production' | 'development';
-
 export async function createStyleStore(onStyleChanged: OnStyleChangedCallback): Promise<IStyleStore> {
   const styleUrl = getStyleUrlAndRemoveItIfNeeded();
   const useStyleUrl = styleUrl && window.confirm("Load style from URL: " + styleUrl + " and discard current changes?");
   let styleStore: IStyleStore;
-  if (__mode__ === 'desktop' && !useStyleUrl) {
+  if ((import.meta as any).env.MODE === 'desktop' && !useStyleUrl) {
```
louwers: Consider using https://vite.dev/guide/env-and-mode.html#intellisense-for-typescript

src/libs/store/style-store-factory.ts
HarelM: Loading a style is async.
The initialization of the store is dependent on weather a style is given in the url or not.
Once a style is loaded there's a need to let the store know if to use the latest from storage or use a new one.
I couldn't find an elegant way to split this logic unfortunately.
Feel free to take a look at the code and let me know if you can think of a better way...

HarelM: Also initializing the ApiStore has network related code, so that is async as well.

src/libs/store/stylestore.ts
HarelM: Sure, I'm a bit confused with all the precommit and other stuff running here, so if you're up to it, please do it.

src/libs/store/style-store-factory.ts
HarelM: Thanks! I looked for it in google and couldn't find it.

HarelM: Fixed in f570804

src/libs/urlopen.ts
HarelM: Done in aded09a

cypress/e2e/maputnik-driver.ts
HarelM: Done in 0a89cef

cypress/e2e/maputnik-driver.ts
```diff
@@ -110,25 +110,25 @@ export class MaputnikDriver {
...
+        url.searchParams.set("style", baseUrl + "geojson-raster-style.json");
         break;
       case "layer":
-        url += `&style=${baseUrl}/example-layer-style.json`;
+        url.searchParams.set("style", baseUrl + "example-layer-style.json");
         break;
       }
       if (zoom) {
-        url += `#${zoom}/41.3805/2.1635`;
+        url.hash = `#${zoom}/41.3805/2.1635`;
```
louwers:
```suggestion
        url.hash = `${zoom}/41.3805/2.1635`;
```

HarelM: This did not fall any tests... Interesting...

PR #1368, opened 2025-09-10
Fix popup issue

HarelM:
## Launch Checklist

When clicking the map to create a popup and then clicking on the map again without closing the popup something breaks in react.
This PR fixes it.

Steps to reproduce:
1. Click on the map
2. Click on the map but not on the popup

The following malformed popup appears:
Before:
image

After - popup looks good.
image

I couldn't make cypress reproduce the exact scenario, but I added some basic tests for popups. 


HarelM:
The problem that appeared in the console was that createRoot was called twice on an element.
I'm not a react expert as well.
But from what I understand the `div` is created once and then called createRoot multiple times.
What the fix here is basically use createRoot once and just call render on it as many times as needed.

I tried other directions, but this was the only one that seems to properly solve it.

The problem is that `close` even from the popup was called after `renderPopup` so there was kind of a race condition...

PR #1375, opened 2025-09-13
Use different icons for different layers

HarelM:
## Launch Checklist

This changes the layers icons a bit in order to make them a bit more distinct:

After:
image

### DIFF
--- a/src/components/IconLayer.tsx
+++ b/src/components/IconLayer.tsx
@@ -1,9 +1,10 @@
 import React from "react";
 
 import type {CSSProperties} from "react";
-import { BsDiamond, BsDiamondFill, BsFonts, BsSun } from "react-icons/bs";
-import { MdBubbleChart, MdOutlineCircle, MdPriorityHigh } from "react-icons/md";
+import { BsDiamondFill, BsFonts, BsSun } from "react-icons/bs";
+import { MdBubbleChart, MdOutlineCircle, MdOutlineStreetview, MdOutlineVerticalShades, MdPriorityHigh } from "react-icons/md";
 import { IoAnalyticsOutline } from "react-icons/io5";
+import { CiMountain1 } from "react-icons/ci";
 
 type IconLayerProps = {
   type: string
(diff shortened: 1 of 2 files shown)


birkskyum (approved):

CommanderStorm:
before we merge, please give me a minute.
I am not entirely happy with a few of the icons (fill + raster + background)

src/components/IconLayer.tsx
```diff
@@ -14,13 +15,13 @@ type IconLayerProps = {
...
+    case "raster": return <BsDiamondFill {...iconProps} />;
     case "hillshade": return <BsSun {...iconProps} />;
-    case "color-relief": return <MdBubbleChart {...iconProps} />;
-    case "heatmap": return <BsDiamond {...iconProps} />;
-    case "fill": return <BsDiamond {...iconProps} />;
-    case "background": return <BsDiamondFill {...iconProps} />;
+    case "color-relief": return <CiMountain1 {...iconProps} />;
+    case "heatmap": return <MdBubbleChart {...iconProps} />;
+    case "fill": return <BsDiamondFill {...iconProps} />;
+    case "background": return <MdOutlineStreetview {...iconProps} />;
```
CommanderStorm: MdOutlineStreetview seems a bit strange for background.
At least I don't consider this very backgroundy 😉

image

How about 

image

src/components/IconLayer.tsx
```diff
@@ -14,13 +15,13 @@ type IconLayerProps = {
 const IconLayer: React.FC<IconLayerProps> = (props) => {
   const iconProps = { style: props.style };
   switch(props.type) {
-    case "fill-extrusion": return <BsDiamondFill {...iconProps} />;
-    case "raster": return <BsDiamond {...iconProps} />;
+    case "fill-extrusion": return <MdOutlineVerticalShades {...iconProps} />;
+    case "raster": return <BsDiamondFill {...iconProps} />;
```
CommanderStorm: raster should likely not be the same as fill.

How about one of these:
imageimageimage

src/components/IconLayer.tsx
```diff
@@ -14,13 +15,13 @@ type IconLayerProps = {
 const IconLayer: React.FC<IconLayerProps> = (props) => {
   const iconProps = { style: props.style };
   switch(props.type) {
-    case "fill-extrusion": return <BsDiamondFill {...iconProps} />;
-    case "raster": return <BsDiamond {...iconProps} />;
+    case "fill-extrusion": return <MdOutlineVerticalShades {...iconProps} />;
```
CommanderStorm: I don't think this icon is very intutive.

image



Do we want to use a building icon instead?

imageimage
image


Alternatively, what do you think about using a cube instead?

image

CommanderStorm (commented):
I have left a few open questions.
Please don't hesitate to overrule me on these, it is design and multiple things usually work here.

HarelM:
I was struggling to find icons that will pass on the relevant message, but I did wanted to make sure that at least they are distinct enough.
Feel free to change it as you see fit. I don't have strong feelings either way...

src/components/IconLayer.tsx
```diff
@@ -14,13 +15,13 @@ type IconLayerProps = {
...
-    case "color-relief": return <MdBubbleChart {...iconProps} />;
-    case "heatmap": return <BsDiamond {...iconProps} />;
-    case "fill": return <BsDiamond {...iconProps} />;
-    case "background": return <BsDiamondFill {...iconProps} />;
+    case "color-relief": return <CiMountain1 {...iconProps} />;
+    case "heatmap": return <MdBubbleChart {...iconProps} />;
+    case "fill": return <BsDiamondFill {...iconProps} />;
+    case "background": return <MdOutlineStreetview {...iconProps} />;
     case "line": return <IoAnalyticsOutline {...iconProps} />;
     case "symbol": return <BsFonts {...iconProps} />;
```
CommanderStorm: 🤔 symbol also still gets the font icon, lets change this too.

This is tough.

What about

image
image
image

HarelM: I'm not sure it the first two will look good since it is small. Location icon is probably as bad as font I guess...
Feel free to use any of the two...

CommanderStorm:
Okay, I edited them as you suggested.
They now look like this

image

I.e:
- no duplicates
- no mixed outline/fill style
- if the symbology is actually better no clue.

HarelM:
LGTM, thanks!

PR #1376, opened 2025-09-13
use biome for formatting

louwers:
## Launch Checklist

 - [ ] Link to related issues.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -22,3 +22,7 @@ repos:
       - id: mixed-line-ending
         args: [ --fix=lf ]
       - id: trailing-whitespace
+  - repo: https://github.com/biomejs/pre-commit
+    rev: v2.2.4 # Use the sha / tag you want to point at
+    hooks:
+      - id: biome-check
--- a/biome.json
+++ b/biome.json
@@ -1,0 +1,6 @@
+{
+  "$schema": "./node_modules/@biomejs/biome/configuration_schema.json",
+  "formatter": {
+    "indentStyle": "space"
+  }
+}
--- a/package-lock.json
+++ b/package-lock.json
@@ -61,8 +61,8 @@
         "url": "^0.11.4"
       },
       "devDependencies": {
+        "@biomejs/biome": "2.2.4",
         "@cypress/code-coverage": "^3.14.5",
-        "@eslint/js": "^9.30.1",
         "@istanbuljs/nyc-config-typescript": "^1.0.2",
         "@rollup/plugin-replace": "^6.0.2",
         "@shellygo/cypress-test-utils": "^5.0.2",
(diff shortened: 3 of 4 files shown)


eslint.config.js
HarelM, line 1: Is this file still needed then?

louwers: Nope!

HarelM:
I've read a bit about this tool, seems like it might need to mature a bit, IDK.
I'm a bit confused on how a linting tool broke the code...?
In any case, I don't like eslint with all the extra plugins and annoying configuration when all I want is basically "regular" coding standard, and this is a project small enough to test it I guess...
So your call... I good with either option...

louwers:
@HarelM ESLint only has third-party formatting plugins now. I like about Biome that it's one package. I don't like jumping on tooling that is immature, but I think it's quite popular, and better in my opinion better than maintaining an ESLint config.

> I'm a bit confused on how a linting tool broke the code...?

That is indeed a bit suspect, but it updated

```ts
          if(!Object.prototype.hasOwnProperty.call(json, "vector_layers")) {
```

to

```
          if (!Object.hasOwn(json, "vector_layers")) {
```

Which TypeScript complained about because its lib option was set to `ES2020` instead of `ES2022`.

It also fixed a few types which caused `// @ts-ignore-error` to fail (because it expected an error).

> this is a project small enough to test it I guess...

I agree. It's pretty fast (written in Rust), documentation seems pretty good, I would say let's give it a shot.


--- 

By the way, I also applied a lot of (semi)-manual fixes to squash all the errors, so I need to verify everything works as intented. Looking into the E2E test fail.

.github/workflows/ci.yml
```diff
@@ -26,10 +26,6 @@ jobs:
           node-version-file: '.nvmrc'
       - run: npm ci
       - run: npm run build
-      - run: npm run lint
```
HarelM, line 29: Should this be repalced with biome or is the pre-commit taking care of this?

build/release-notes.js
```diff
@@ -21,12 +21,11 @@ const changelog = fs.readFileSync(changelogPath, "utf8");
 const regex = /^## (\d+\.\d+\.\d+.*?)\n(.+?)(?=\n^## \d+\.\d+\.\d+.*?\n)/gms;
 
 const releaseNotes = [];
-let match;
-// eslint-disable-next-line no-cond-assign
-while (match = regex.exec(changelog)) {
+// Iterate over all matches without using assignment in the loop condition
```
HarelM, line 24: I hope this is the same, I'm not sure I have a good way of testing it...

cypress/e2e/layers.cy.ts
```diff
@@ -729,7 +769,10 @@ describe("layers", () => {
         type: "background",
       });
 
-      when.dragAndDropWithWait("layer-list-item:" + firstId, "layer-list-item:" + thirdId);
+      when.dragAndDropWithWait(
+        `layer-list-item:${firstId}`,
```
HarelM, line 773: I can't say these extra lines are really helpful when reading this code... But I guess it is what it is...

cypress/e2e/modal-driver.ts
```diff
@@ -8,9 +8,9 @@ export default class ModalDriver {
     fillLayers: (opts: { type: string; layer?: string; id?: string }) => {
       // Having logic in test code is an anti pattern.
       // This should be splitted to multiple single responsibility functions
-      const type = opts.type;
+      const type: string = opts.type;
```
HarelM, line 11: I don't think this is needed in terms of typescript definitions (as opposed to two lines below where it is needed).

HarelM:
I didn't go over all the files, If there are specific things you want me to review let me know, otherwise if the rest was done automatically, there's no need for my review.

HarelM (approved):

.github/workflows/ci.yml
louwers: Yes pre-commit does it now, which does auto fixes for formatting.

birkskyum:
Thanks for looking into speeding this up! I think biome is good for now, but in the long run I'd like to consolidate our typescript projects on the vite ecosystem, which we've moved towards for maplibre-style-spec and maplibre-gl-js, where the oxc tools (oxlint / oxfmt) and tsgo (can be used with tsgolint) will be central quite soon.

HarelM:
@louwers what are we planning for this PR?

louwers:
I can try out oxlint instead since @birkskyum mentioned he prefers it.

Does not make sense to switch to another formatter in a year from now.

PR #1379, opened 2025-09-14
Add globe support in Maputnik UI

HarelM:
## Launch Checklist

Add a small drop down to select mercator or globe.
This isn't a fully covered field as one can set an expression there, but I believe this is good enough for most cases.

Before:
image
After:
image

### DIFF
--- a/src/components/modals/ModalSettings.tsx
+++ b/src/components/modals/ModalSettings.tsx
@@ -1,6 +1,6 @@
 import React from "react";
 import latest from "@maplibre/maplibre-gl-style-spec/dist/latest.json";
-import type {LightSpecification, StyleSpecification, TerrainSpecification, TransitionSpecification} from "maplibre-gl";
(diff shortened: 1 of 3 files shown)


birkskyum (approved):

src/components/modals/ModalSettings.tsx
```diff
@@ -240,39 +259,51 @@ class ModalSettingsInternal extends React.Component<ModalSettingsInternalProps>
...
-          onChange={this.changeTransitionProperty.bind(this, "duration")}
+          onChange={(value) => this.changeTransitionProperty("duration", value)}
+        />
+
+        <FieldSelect
+          label={t("Projection")}
+          data-wd-key="modal:settings.projection"
+          options={[
+            ["", "Undefined"],
+            ["mercator", "Mercator"],
```
birkskyum, line 302: can we have the vertical-perspective option here as well?

HarelM: I added it, but I wasn't not sure it was interesting enough to add, so I removed it.
Feel free to add it if you think it's valuable.

PR #1380, opened 2025-09-14
Fix firefox ability to load file, add firefox tests back

HarelM:
## Launch Checklist

This PR fixes the problem with firefox not able to load a file when clicking the open file button.
It does so by removing a very old dependency: `react-file-reader-input`
It uses the "regular" `<label><a/><input type="file" /><label>` approach.
It adds tests for both firefox and chrome although I'm not sure the firefox tests will stay... 


CommanderStorm (approved):
Nice

PR #1386, opened 2025-09-16
Codemirror 5 to 6 upgrade

### LINKED ISSUE #891: Upgrade codemirror
Maputnik currently uses codemirror 5, but codemirror 6 is avaiable, but they have changed the API dramatically.
Some of it is explained here:
https://codemirror.net/docs/migration/
But there are a lot of things that needs to be defined and updated and I'm not familiar enough with the original definitions and requirements in order to "safely" update this package.
If anyone is familiar with codemirror enough to easily update the package it may be helpful.

HarelM:
## Launch Checklist

- Resolves #891

This PR upgrades code mirror from version 5 to version 6.
It should not change any functionality dramatically.
The filter and other expressions have line numbers now as I was not able to remove those without introducing a lot of code, which I preferred not to.

Before:

image

After:
image

 - [x] Link to related issues.


src/components/FilterEditor.tsx
```diff
@@ -293,7 +294,7 @@ class FilterEditorInternal extends React.Component<FilterEditorInternalProps, Fi
               this.props.onChange(defaultFilter);
             }}
             fieldName="filter"
-            fieldSpec={fieldSpec}
+            fieldSpec={fieldSpec as any}
```
CommanderStorm: is this intentional?
The type of `fieldSpec` was also changed

src/components/FilterEditor.tsx
```diff
@@ -3,7 +3,8 @@ import { TbMathFunction } from "react-icons/tb";
 import { PiListPlusBold } from "react-icons/pi";
```
CommanderStorm, line 3: css bug:
Filter editors don't get the correct "Text"-cursor and thus feel weird

src/libs/codemirror-mgl.ts
```diff
@@ -1,174 +1,219 @@
...
+  return new EditorView({
+    doc: props.value,
+    extensions: [
+      basicSetup,
+      json(),
+      oneDark,
+      new Compartment().of(EditorState.tabSize.of(2)),
+      EditorView.theme({
+        "&": {
+          fontSize: "9pt"
```
CommanderStorm: This is more of a global issue, but I think we should increase the size of fonts in maputnik.
9pt is a bit small, at least for my eyes.

CommanderStorm (commented):
That is a lot of change.

I have two comments, but we can also split them out into separate PRs:

The spacing here seems a little high

image

And I would like to see the cursor be the text cursor for `FilterEditor`

src/libs/codemirror-mgl.ts
```diff
@@ -1,174 +1,219 @@
...
+        }
+        if (update.focusChanged) {
+          if (update.view.hasFocus) {
+            props.onFocus();
+          } else {
+            props.onBlur();
+          }
+        }
+      }),
+      lintGutter(),
```
CommanderStorm: This seems to be the issue that I was seeing above.

```suggestion
```

src/libs/codemirror-mgl.ts
HarelM: I agree, I mainly wanted to keep the appearance similar to the current implantation. I think a size overhaul is outside the scope of this PR.

src/components/FilterEditor.tsx
HarelM: The type of filed spec represent something correct now, this component creates something that is partial and thus needs this hack. 
I'll improve it to avoid the need for "as any".

src/libs/codemirror-mgl.ts
HarelM: This allows to place the error in the gutters, which I think is a good UX.
maybe folding is not needed. 
I didn't want to customize the basicSetup too much, but that's possible too if you think it's worth the extra complexity.

HarelM:
> A function like getArrayPositionalFromAst could probably use a unit test.

You are right, but this requires introducing a unit test framework, which is currently missing. 

This specific function didn't change, BTW, I just added typings to it and some minor "return" refactoring...

src/components/FilterEditor.tsx
HarelM: This was fixes in 08880a791f96b45501ffd3fa540a7a169a7636d6
Apparently this is not used and can be passed as an empty object.

src/components/FilterEditor.tsx
HarelM: I'm not sure why that is actually, they are created the same way basically...

HarelM: I don't want to add a custom css just for that to be honest, but if you feel it's worth it I'll do it...

HarelM: Fixed in 7459379e92cc6a267d3fb6e2a6af74a0a821726b

HarelM:
I've added unit tests using vitest and made sure the coverage of the ast related method is 100%, thanks for the nudge!
This is ready for review/merge from my side.

HarelM:
@louwers, @CommanderStorm, @birkskyum anything else that is needed in order to get this merged?

louwers:
Missed your comment, looks good!

PR #1426, opened 2025-10-05
Add code editor for maputnik

### LINKED ISSUE #820: Having a json panel with the complete style
For each layer we can check the json created at the bottom of the style panel.
It could be great to have a json panel in where we can see the whole json of the style.

We can change values in the json and it will update the parameters in the corresponding form.
We can import and export json from this panel by copy/paste.

HarelM:
## Launch Checklist

This PR adds the ability to look at the entire style and edit it in a code editor that supports syntax highlight, errors, search and more.

- Resolves #820

CC: @Kanahiro as I know you did something similar, probably has better performance...

After:
image

 - [x] Link to related issues.


src/components/AppToolbar.tsx
```diff
@@ -221,23 +222,27 @@ class AppToolbarInternal extends React.Component<AppToolbarInternalProps> {
           </a>
         </div>
         <div className="maputnik-toolbar__actions" role="navigation" aria-label="Toolbar">
-          <ToolbarAction wdKey="nav:open" onClick={this.props.onToggleModal.bind(this, "open")}>
+          <ToolbarAction wdKey="nav:code-editor" onClick={() => this.props.onToggleModal("codeEditor")}>
+            <MdCode />
+            <IconText>{t("Code Editor")}</IconText>
+          </ToolbarAction>
```
CommanderStorm: Users commonly click the first action first.

Since If the code editor is empty this is not very actionable, could we either move this after the open toolbar item or hide unless the action makes sense?

src/locales/de/translation.json
```diff
@@ -197,6 +205,7 @@
   "Public access token for LocationIQ services.": "Öffentlicher Zugriffstoken für LocationIQ-Dienste.",
   "Style Renderer": "Stil-Renderer",
   "Choose the default Maputnik renderer for this style.": "Wähle den Standard-Renderer für diesen Stil aus.",
+  "Press <1>ESC</1> to lose focus": "Drück <1>ESC</1>, um den Fokus zu verlieren",
```
CommanderStorm: Is this translation still correct? (I was looking into this as the german wording is akward)

https://github.com/maplibre/maputnik/blob/454d8d8b108b5a14c3a5186953b97ea2a0fee6d3/src/components/modals/ModalShortcuts.tsx#L113

src/locales/de/translation.json
```diff
@@ -124,6 +131,7 @@
   "Terrain exaggeration": "Gelände-Übertreibung",
```
CommanderStorm: nit/unrelated: Verrry literal translations into german are .. akward.

```suggestion
  "Terrain exaggeration": "Geländeüberhöhung",
```

CommanderStorm (changes requested):
Looks and works mostly well.

What I don't think users will get is how to exit this mode. I think we need something like this:

image

Otherwise mostly LGTM

src/locales/de/translation.json
HarelM: This one just moved, don't ask me why i18n doesn't sort the file by key instead...

src/components/AppToolbar.tsx
HarelM: It is acceptable actionable if you paste the style into it, but yeah, I can move this forward, I don't have strong feelings for this item, although I guess we should use a menu instead of all these buttons at the top...
But that's a different story.

src/locales/de/translation.json
HarelM: I added the relevant translation use co-pilot, so I don't know if any of them are good, but if it is bad enough maybe someone will help fix it.

HarelM:
I've updated the images with the recent fixes.

CommanderStorm (approved):
Much better.
Lets just make the button slightly bigger and have better contrast.

Currently, I don't think users will find this due to the missing contrast and tiny size.

image

Afterwards, LGTM!

HarelM:
I don't want to invent a different styling for a button, so I used the button style, since it's always at the top I'm sure users will find it if they would need to.
Also clicking the original button that opened the editor will go back to the layers list.
If someone will report this as an issue I'll be happy to help, but I want to see that people are reporting issues as I'm not sure how many active users this website has, to be honest...

PR #1431, opened 2025-10-06
chore(ci): reduce dependency churn

CommanderStorm:
Currently, dependabot very much churns through PRs on this repo.
This is particularly spammy for people watching this repo like me.

This PR moves the interval to be one PR weekly, with collodown for non-maplibre releases.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,10 +8,39 @@ updates:
   - package-ecosystem: "npm" # See documentation for possible values
     directory: "/" # Location of package manifests
     schedule:
-      interval: "daily"
+      interval: "weekly"
     open-pull-requests-limit: 20
     versioning-strategy: increase
+    groups:
+      all-actions-version-updates:
+        applies-to: version-updates
+        patterns: [ '*' ]
+      all-actions-security-updates:
+        applies-to: security-updates
+        patterns: [ '*' ]
+    cooldown:
+      default-days: 5
+      semver-major-days: 5
+      semver-minor-days: 3
+      semver-patch-days: 3
+      include:
+        - "*"
+      exclude:
+        - "@maplibre/*"
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
-      interval: "daily"
+      interval: "weekly"
+    groups:
+      all-actions-version-updates:
+        applies-to: version-updates
+        patterns: [ '*' ]
+      all-actions-security-updates:
+        applies-to: security-updates
+        patterns: [ '*' ]
+    cooldown:
+      default-days: 3
+      # no semver support for github-actions
+      # => no specific configuration for this
+      include:
+        - "*"


nyurik:
i think @HarelM was pretty much against grouping it - probably due to it not working so well for javascript, but definitly lets make it less frequent

.github/dependabot.yml
```diff
@@ -8,10 +8,39 @@ updates:
...
+      interval: "weekly"
     open-pull-requests-limit: 20
     versioning-strategy: increase
+    groups:
+      all-actions-version-updates:
+        applies-to: version-updates
+        patterns: [ '*' ]
+      all-actions-security-updates:
+        applies-to: security-updates
+        patterns: [ '*' ]
```
CommanderStorm:
```suggestion
```

.github/dependabot.yml
```diff
@@ -8,10 +8,39 @@ updates:
...
     schedule:
-      interval: "daily"
+      interval: "weekly"
+    groups:
+      all-actions-version-updates:
+        applies-to: version-updates
+        patterns: [ '*' ]
+      all-actions-security-updates:
+        applies-to: security-updates
+        patterns: [ '*' ]
```
CommanderStorm:
```suggestion
```

CommanderStorm:
I defer to @HarelM if he wants monthly or weekly.

HarelM:
I'm deleting mails from dependabot, and I think you can configure some rules to avoid spam in terms of notifications to the mail.
This PR will only moved the notification to happen once a week, but you'd still get all of them unless a package published two releases within a week, which usually doesn't happen. 
And since I'm against grouping I'm not entirely sure what this will solve TBH...

.github/dependabot.yml
```diff
@@ -8,10 +8,25 @@ updates:
   - package-ecosystem: "npm" # See documentation for possible values
     directory: "/" # Location of package manifests
     schedule:
-      interval: "daily"
+      interval: "weekly"
```
CommanderStorm:
```suggestion
      interval: "daily"
```

.github/dependabot.yml
```diff
@@ -8,10 +8,25 @@ updates:
...
+      semver-patch-days: 3
+      include:
+        - "*"
+      exclude:
+        - "@maplibre/*"
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
-      interval: "daily"
+      interval: "weekly"
```
CommanderStorm:
```suggestion
      interval: "daily"
```

CommanderStorm:
Fair. Lets then just add the cooldown then.

I have issues with working email based, that misses updates, so that workflow does not work for me.

nyurik:
how about we simply lower it to monthly?

HarelM:
I prefer more raid updates than once a month.

PR #1455, opened 2025-10-17
fix: remove broken max width from Expression properties

CommanderStorm:
Before vs after should be self-explanatory what motivated me to remove the max-width.

image

Resolves https://github.com/maplibre/maputnik/issues/1445


HarelM (approved):

HarelM:
THANKS! I should've removed it but forgot...

PR #1485, opened 2025-11-01
fix: keep headers visible when scrolling left panes

jeresuikkila:
Keeps headers visible when scrolling left panes as described in the [issue 951
](https://github.com/maplibre/maputnik/issues/951)

The fix was manually confirmed to be working, see video below.

## Before (taken from the issue)
image

## After
https://github.com/user-attachments/assets/ab5e4a6f-c5f9-44fd-850d-8eac58c35c68

### DIFF
--- a/cypress/e2e/layer-editor.cy.ts
+++ b/cypress/e2e/layer-editor.cy.ts
@@ -266,37 +266,15 @@ describe("layer editor", () => {
         layer: "example",
       });
 
-      // Wait for properties to render
       when.wait(500);
-
-      // Get the layer editor header
-      const header = get.element(".maputnik-layer-editor > header");
-
-      // Verify header is initially visible
+      const header = get.elementByTestId("layer-editor.header");
       then(header).shouldBeVisible();
 
-      // Verify header has sticky positioning and proper z-index
-      header.should("have.css", "position", "sticky");
-      header.should("have.css", "top", "0px");
-      header.should("have.css", "z-index", "10");
-
-      // Scroll the layer editor container (use ensureScrollable: false to avoid flakiness)
       get.element(".maputnik-scroll-container").scrollTo("bottom", { ensureScrollable: false });
       when.wait(200);
 
-      // Header should still be visible after scrolling
       then(header).shouldBeVisible();
-
-      // Verify layer menu button in header is still accessible
       then(get.elementByTestId("skip-target-layer-editor")).shouldBeVisible();
-    });
-
-    it("should have opaque background on header", () => {
-      when.modal.fillLayers({
-        type: "background",
-      });
-      const header = get.element(".maputnik-layer-editor > header");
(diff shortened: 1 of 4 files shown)


HarelM:
Thanks!
Can you check why the end to end tests are failing?
I think I've sent a retry and it still didn't pass...

HarelM:
Is it possible to add an end-to-end test for this new functionality,I know it's only css, but still, no tests means it will break in the future...

jeresuikkila:
Yeah I'll fix the tests

cypress/e2e/layer-editor.cy.ts
```diff
@@ -257,4 +257,46 @@ describe("layer editor", () => {
...
+      when.modal.fillLayers({
+        type: "symbol",
+        layer: "example",
+      });
+
+      // Wait for properties to render
+      when.wait(500);
+
+      // Get the layer editor header
+      const header = get.element(".maputnik-layer-editor > header");
```
HarelM: Can we use `data-wd-key` instead of css?

cypress/e2e/layer-editor.cy.ts
```diff
@@ -257,4 +257,46 @@ describe("layer editor", () => {
...
+
+      // Verify header is initially visible
+      then(header).shouldBeVisible();
+
+      // Verify header has sticky positioning and proper z-index
```
HarelM: I would consider removing those CSS check and focus on behavior.

cypress/e2e/layer-editor.cy.ts
```diff
@@ -257,4 +257,46 @@ describe("layer editor", () => {
...
+      when.wait(200);
+
+      // Header should still be visible after scrolling
+      then(header).shouldBeVisible();
+
+      // Verify layer menu button in header is still accessible
+      then(get.elementByTestId("skip-target-layer-editor")).shouldBeVisible();
+    });
+
+    it("should have opaque background on header", () => {
```
HarelM: Consider removing this test.

cypress/e2e/layers-list.cy.ts
```diff
@@ -515,4 +515,46 @@ describe("layers list", () => {
...
+      // Wait for layers to render
+      when.wait(500);
+
+      // Get the layer list header
+      const header = get.element(".maputnik-layer-list-header");
+
+      // Verify header is initially visible
+      then(header).shouldBeVisible();
+
+      // Verify header has sticky positioning and proper z-index
```
HarelM: Same notes as the other test.

HarelM:
Thanks for adding the tests!
I've added some minor comment, otherwise looks good!

jeresuikkila:
Thanks for your patience! This removes the extraneous tests, focuses on behavior and uses test ids.

PR #1486, opened 2025-11-02
chore(deps): update vitest and group them

HarelM:
## Launch Checklist

Update vitest packages and group them in dependabot.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -118,9 +118,9 @@
       }
     },
     "node_modules/@angular/common": {
-      "version": "20.3.0",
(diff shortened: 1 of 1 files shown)


birkskyum:
I think vitest v4 might have quite a few migration gotchas
- https://vitest.dev/guide/migration.html#vitest-4

HarelM:
That's true, but in this repo there are like 5 vitests, so I don't expect pain here.
I just finished upgrading maplibre-gl-js to vitest 4, it was a lot more painful...

PR #1549, opened 2025-12-01
allow root relative paths

lukasalexanderweber:
## Launch Checklist

when launching maputnik with `--static /data` those files are available under e.g. `localhost:8000/static`. In this case, root relative urls are enough (e.g. requsting `/static/fonts`...). This does currently lead to a validation error:

grafik

after this change it works:

grafik

### DIFF
--- a/src/libs/urlopen.ts
+++ b/src/libs/urlopen.ts
@@ -45,8 +45,8 @@ function getProtocolSafe(url: string): { protocol?: string, isLocal?: boolean } 
   }
 };
 
-export function validate(url: string): ErrorType {
-  if (url === "") {
+export function validate(url?: string): ErrorType {
+  if (!url) {
     return ErrorType.None;
   }
 


HarelM:
Thanks for taking the time to open this PR!
Wouldn't this allow creating invalid styles though?

lukasalexanderweber:
@HarelM according to the [spec](https://docs.mapbox.com/style-spec/reference/sources/#raster-url) they would be invalid.

But both openlayers and maplibre gl js can perfectly work with root-relative paths and will just add their base path in front, providing the correct protocol.

It makes style development easier when hosting them from different domains.

Imagine

`"glyphs": "/fonts/mapbox/{fontstack}/{range}.pbf"`

and

`https://devserver.com`, `https://integrationserver.com`, `https://productionserver.com`

when loading the glyphs within a map hosted on `https://devserver.com` they will just be resolved to 
`https://devserver.com/fonts/mapbox/{fontstack}/{range}.pbf`. Same for the others.

if I don't use root-relative paths I'd need to adjust glyphs in every environment, which is pretty anoying. Additionally, we have some Cloudflare magic that `https://devserver.lan` + `https://devserver.com` lead to the same services and data (but with different auths), so working with "absolute" paths wont even work anymore.

Edit: If you don't want this in your codebase we can just keep the changes in a fork, I just thought it might make sense to offer this adjustment to you

HarelM (approved):

HarelM:
If this is a valid use case that works, then sure, why not.

HarelM:
It might be that e2e is failing due to this fix, I saw an error of "Cannot read properties of undefined (reading 'startsWith')", so there might be a missing "?" to avoid the case of undefined string?

lukasalexanderweber:
@HarelM here is a [quick fix](https://github.com/maplibre/maputnik/pull/1549/commits/2464da6b7f1f09f65decd0158055fc3b1f4a9b37)

PR #1552, opened 2025-12-02
set correct map view if opened stylefile provides a map view and the current map is empty

lukasalexanderweber:
## Launch Checklist

closes https://github.com/maplibre/maputnik/issues/1546

 - [x] Link to related issues.
 https://github.com/maplibre/maputnik/issues/1546


HarelM:
It's better to have the conversation here if it is related to the PR details and not the issue's problem statement. 
Can you clarify which problem you are facing?

lukasalexanderweber:
I'd like to test that the url is correct and that loading a second style file does not alter the maps zoom and center. But the commented out tests in cypress/e2e/map.cy.ts do not work and I'd like a second opinion where to go from there

HarelM:
It might be that it takes time for the url to change. 
I would recommend running the tests locally using npm run cy:open and adding it.only to the test to run only your test and see what happens in the browser when the test run, and continue debugging from there. 
This is not maputnik specific but Cypress related and you can consult with some AI or do a Google search.

lukasalexanderweber:
wow `cy:open` is a great tool, fun to work with. I thought it might be related to the MaputnikDriver but it just sits on top of cypress. I've also added a fix to the popup test, which should now be stable

cypress/e2e/map.cy.ts
```diff
@@ -22,6 +22,27 @@ describe("map", () => {
...
       );
     });
+
+    it("via style file definition", () => {
+      when.setStyle("zoom_center");
+      then(get.elementByTestId("maplibre:ctrl-zoom")).shouldBeVisible();
+      then(get.elementByTestId("maplibre:ctrl-zoom")).shouldContainText(
+        "Zoom: " + (7)
+      );
+      cy.location("hash").should("contain", "#7/51.5/0");
```
HarelM: `cy` calls should be kept in the driver and not in the tests themselves.

src/components/App.tsx
```diff
@@ -466,15 +475,25 @@ export default class App extends React.Component<any, AppState> {
...
     this.setState({
       mapStyle: newStyle,
       dirtyMapStyle: dirtyMapStyle,
+      mapView: isEmptyStyle && zoom && center ? {
+        zoom: zoom,
+        center: {
+          lng: center[0],
+          lat: center[1],
+        },
+        _from: "app"
```
HarelM, line 490: Are you sure this is the best way to achieve this? I feel like this is a hack to use the state for something we shouldn't be using it for.
Is there a different way to achieve this functionality?

lukasalexanderweber: I only found this way. We need a marker to not run into an Infinite recursion

cypress/e2e/map.cy.ts
lukasalexanderweber: does it make sense to wrap all cy functionality in the driver? Are there advantages?

HarelM: Yes, the idea is that the code in the tests can be read by someone who is not a developer (product marketing if you will), it should describe what happens using "get", "when" and "then".
In the driver you can create methods with this terminology.
Also in theory (but probably not in practice) you should be able to change the underlying infrastructure from cypress to something else without needing to rewrite all the tests.

lukasalexanderweber: makes sense 👍  I'll update the tests tomorrow

cypress/fixtures/example-style-with-zoom-7-and-center-0-51.json
HarelM, line 1: Can you simplify this file to have minial data?

cypress/fixtures/example-style-with-zoom-5-and-center-50-50.json
HarelM, line 1: Is this file still needed?

src/components/App.tsx
HarelM: I'm wondering if there might be a way maybe to split the style setting logic instead of this flag...?

cypress/fixtures/example-style-with-zoom-5-and-center-50-50.json
lukasalexanderweber: Yes, it contains different Zoom and Center to test that opening a second stylefile does not move the map again

src/components/App.tsx
lukasalexanderweber: but we need to keep the map views state consistent

cypress/e2e/map.cy.ts
lukasalexanderweber: this is now abstracted

cypress/fixtures/example-style-with-zoom-7-and-center-0-51.json
lukasalexanderweber: done

lukasalexanderweber:
@HarelM friendly reminder that I've incorporated your review suggestions, this should be ready

cypress/e2e/maputnik-driver.ts
```diff
@@ -23,6 +23,13 @@ export class MaputnikAssertable<T> extends Assertable<T> {
         then(this.chainable).shouldDeepNestedInclude(style);
       })
     );
+
+  shouldHaveUrlHashContaining = (expectedHash: string) =>
```
HarelM: Is this in use? I prefer that over cypress' `should("contain", ...` usage.

cypress/e2e/maputnik-driver.ts
```diff
@@ -92,6 +99,20 @@ export class MaputnikDriver {
...
+      this.helper.given.interceptAndMockResponse({
+        method: "GET",
+        url: baseUrl + "example-style-with-zoom-and-center.json",
+        response: {
+          fixture: "example-style-with-zoom-and-center.json",
+        },
+      });
+      this.helper.given.interceptAndMockResponse({
+        method: "GET",
+        url: baseUrl + "example-style-with-zoom-and-center2.json",
```
HarelM: Maybe a better name to indicate what's in it? Maybe have the zoom as part of the file name to differentiate them?

src/components/App.tsx
HarelM: I don't have a better idea how to solve it, so I guess we'll keep this implementation...

HarelM:
I'm hoping some of the functionality needed here can be moved to the following library:
- https://github.com/ShellyDCMS/cypress-test-utils/issues/211

But I don't think we should wait with this PR for this.
Let's make it so that it will be easy to pass the methods from here to there when needed and follow the patterns that are used there.

cypress/e2e/maputnik-driver.ts
lukasalexanderweber: no I've deleted it

cypress/e2e/maputnik-driver.ts
lukasalexanderweber: good idea, I've changed the file names

lukasalexanderweber:
you mean this [extension](https://github.com/lukasalexanderweber/maputnik/blob/update_map_view/cypress/e2e/maputnik-cypress-helper.ts#L13) of the `get` might be added in the external lib?

```
public get = {
  locationHash: (): Cypress.Chainable<string> => cy.location("hash"),
  ...this.helper.get,
};
```

HarelM (approved):

PR #1596, opened 2025-12-17
Added Korean language support

jinifor:
Added Korean language support.

image

To ensure accurate meaning, some terms were written following the criteria below.

- Map → '맵' : 'Map' is rendered as '맵', which is more commonly used than '지도' in GIS tools.
- Stops → '기준점' : Transliterating the term as '스톱' does not clearly convey its meaning. Since it represents a reference point used for interpolation, it is translated as '기준점'.
- Light → '라이트' : Literal translations such as '광원', '빛' or '조명' awkward in a 3D rendering context, so the term is transliterated.
- Focus → '포커스' : The literal translation '초점' sounds awkward in a UI context, so it has been transliterated for clarity.
- Terrain → '터레인' : In the context of 3D maps, '터레인' is a more commonly used term than '지형'.
- Paint → '페인트' : The term is transliterated to preserve its meaning as defined in the Style Specification.


HarelM (approved):
THANKS!
I'll ping you from time to time when new text is added. I'll probably be using some LLM for initial translation, but feel free to open PRs to fix stuff that wasn't translated correctly.

PR #1643, opened 2026-01-19
fix root relative paths

lukasalexanderweber:
## Launch Checklist

this is a follow up to https://github.com/maplibre/maputnik/pull/1549

when testing this, I only used raster tiles. However root relative vector tiles are not loaded correctly by maplibre gl js: https://github.com/maplibre/maplibre-gl-js/issues/6818

The current state is problematic as with https://github.com/maplibre/maputnik/pull/1549 we don't have the visuals that the url is wrong.

Till [the maplibre issue](https://github.com/maplibre/maplibre-gl-js/issues/6818) is fixed (if it even need to be fixed), we can make use of [transformRequest](https://maplibre.org/maplibre-gl-js/docs/API/type-aliases/MapOptions/#transformrequest) where we convert root relative requests to "real" requests

 - [x] Link to related issues.
 - [x] Include before/after visuals or gifs if this PR includes visual changes -> in maplibre ticket

### DIFF
--- a/src/components/MapMaplibreGl.tsx
+++ b/src/components/MapMaplibreGl.tsx
@@ -143,7 +143,7 @@ class MapMaplibreGlInternal extends React.Component<MapMaplibreGlInternalProps, 
       // not support this for everything:
       // https://github.com/maplibre/maplibre-gl-js/issues/6818
       transformRequest: (url) => {
-        if (url.startsWith('/')) {
+        if (url.startsWith("/")) {
           url = `${window.location.origin}${url}`;
         }
         return { url };


lukasalexanderweber:
there is a Map created in https://github.com/maplibre/maputnik/blob/main/src/components/modals/ModalExport.tsx#L72 IDK what this does and if it needs support for this

HarelM (approved):
Thanks for taking care of this!
Can you add a small changelog entry for this change?

HarelM:
> there is a Map created in https://github.com/maplibre/maputnik/blob/main/src/components/modals/ModalExport.tsx#L72 IDK what this does and if it needs support for this

This export and html file with the style built into it, I'm not sure this feature is actually used, so I won't worry about it.
If anyone complains then he can use absolute paths instead, it will probably be a "more correct" solution...

lukasalexanderweber:
As nothing was released yet I thought https://github.com/maplibre/maputnik/blob/6093ec9047b00d9ed7bacdd584d0884a3a4f3105/CHANGELOG.md?plain=1#L15 is enough

I could change "Allow" to "Support"

HarelM:
Fair enough.

PR #1680, opened 2026-03-02
fix: resolve layer click selection conflict with dnd-kit drag listeners

sank64:
## Summary

Fixes an issue where clicking on layer names in the layer list doesn't reliably open the layer properties editor.

## Problem

The `DraggableLabel` component in `LayerListItem.tsx` spreads dnd-kit's `dragListeners` (which include `onPointerDown`) onto the layer name handle div. This pointer capture can intercept normal click events, preventing the parent `<li>`'s `onClick` handler from firing. The result is that clicking a layer name sometimes fails to select and display its properties.

## Solution

Added an explicit `onClick` handler to the `DraggableLabel` component that:
1. Calls `e.stopPropagation()` to prevent event bubbling conflicts
2. Directly invokes the `onSelect` callback passed from the parent

The `onClick` is spread **after** `dragListeners`, ensuring it takes precedence. This approach preserves full drag-and-drop functionality while guaranteeing click selection always works.

## Changes

- **`src/components/LayerListItem.tsx`**: Added `onSelect` prop to `DraggableLabel`, implemented `handleClick` with `stopPropagation`, applied `onClick` after drag listeners on the handle div.

### DIFF
--- a/src/components/LayerListItem.tsx
+++ b/src/components/LayerListItem.tsx
@@ -13,11 +13,19 @@ type DraggableLabelProps = {
   layerType: string
   dragAttributes?: React.HTMLAttributes<HTMLElement>
   dragListeners?: React.HTMLAttributes<HTMLElement>
+  onSelect?: () => void
 };
 
 const DraggableLabel: React.FC<DraggableLabelProps> = (props) => {
   const {dragAttributes, dragListeners} = props;
-  return <div className="maputnik-layer-list-item-handle" {...dragAttributes} {...dragListeners}>
+
+  const handleClick = (e: React.MouseEvent) => {
+    // Ensure layer selection fires even when dnd-kit captures the pointer
+    e.stopPropagation();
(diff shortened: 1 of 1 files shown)


HarelM:
I'm guessing the tests failures are related to changes to chrome and webgl... :-/
It would be great if you could take care of it, otherwise I'll need to find some time to update all the dependencies here and fix the webgl stuff...

HarelM:
A different solution I implemented in a different place was to add a "drag" handle next the title, but if this solution works well then that's fine too.

sank64:
> I'm guessing the tests failures are related to changes to chrome and webgl... :-/ It would be great if you could take care of it, otherwise I'll need to find some time to update all the dependencies here and fix the webgl stuff...

I have tried to fix them, but they seem unrelated to my edit. 

I have published a new PR https://github.com/maplibre/maputnik/pull/1682 with all the fixes I could make. You can close this PR

PR #1682, opened 2026-03-02
feat: add adjustable sidebar width with drag handles

sank64:
PR 1682 - preview

## Summary
Adds draggable resize handles to the sidebar, allowing users to adjust both the total sidebar width and the layer list / properties drawer split. Also fixes a bug where clicking on layer name text did not open properties.
Address all corrections required in https://github.com/maplibre/maputnik/pull/1677

## Changes
- Convert AppLayout from class to functional component
- Add outer resize handle between sidebar and map
- Add inner resize handle between layer list and properties drawer
- Use CSS custom properties for dynamic widths
- Persist sidebar width and list/drawer ratio to localStorage
- Fix: clicking layer name text now opens layer properties (dnd-kit dragListeners were intercepting clicks)
- Extract sidebar helpers to `src/libs/sidebar.ts`
- Add unit tests for sidebar helpers
- Add Cypress e2e test for resize handles

Note: CI e2e failures are all `Failed to initialize WebGL` - a Chrome 145 headless GPU issue on GitHub Actions, unrelated to this PR.


cypress/e2e/sidebar-resize.cy.ts
```diff
@@ -0,0 +1,50 @@
...
+describe("sidebar resize", () => {
+  const { beforeAndAfter, when } = new MaputnikDriver();
+  beforeAndAfter();
+
+  beforeEach(() => {
+    when.setStyle("both");
+  });
+
+  it("resize handle is visible", () => {
+    cy.get("[data-testid='sidebar-resize-handle']").should("exist").and("be.visible");
```
HarelM: This file should be using the driver and helper instead of using cy methods directly.
You don't need to close the PR and open a new one but simply push more commit to the branch you are working on.

src/components/AppLayout.tsx
```diff
@@ -1,54 +1,182 @@
...
+  MIN_DRAWER_WIDTH,
+  clampSidebarWidth,
+  getSavedSidebarWidth,
+  getSavedListRatio,
+  saveSidebarWidth,
+  saveListRatio,
+} from "../libs/sidebar";
 
-type AppLayoutInternalProps = {
+type AppLayoutProps = {
```
HarelM, line 17: Why is this not using `WithTranslation` anymore?

src/components/AppLayout.tsx
```diff
@@ -1,54 +1,182 @@
...
+        }
+        {!props.codeEditor && <>
+          <div className="maputnik-layout-list">
+            {props.layerList}
+          </div>
+          <div
+            className="maputnik-layout-resize-handle maputnik-layout-resize-handle--inner"
+            data-testid="inner-resize-handle"
+            onMouseDown={handleInnerMouseDown}
+            title="Drag to resize list / editor split"
```
HarelM: This probably needs translation.

src/components/AppLayout.tsx
```diff
@@ -1,54 +1,182 @@
...
-        {this.props.bottom && <div className="maputnik-layout-bottom">
-          {this.props.bottom}
-        </div>
-        }
-        {this.props.modals}
+          <div
+            className="maputnik-layout-resize-handle"
+            data-testid="sidebar-resize-handle"
+            onMouseDown={handleMouseDown}
+            title="Drag to resize sidebar"
```
HarelM: This probably needs translation too.

src/components/AppLayout.tsx
```diff
@@ -1,54 +1,182 @@
...
+
+    document.addEventListener("mousemove", handleMouseMove);
+    document.addEventListener("mouseup", handleMouseUp);
+    return () => {
+      document.removeEventListener("mousemove", handleMouseMove);
+      document.removeEventListener("mouseup", handleMouseUp);
+    };
+  }, []);
+
+  const layoutStyle = {
```
HarelM: Since this is used below, it might be better to inline it.

src/components/AppLayout.tsx
```diff
@@ -1,54 +1,182 @@
...
+        document.body.style.cursor = "";
+        document.body.style.userSelect = "";
+        setListRatio((r) => {
+          saveListRatio(r);
+          return r;
+        });
+      }
+    };
+
+    document.addEventListener("mousemove", handleMouseMove);
```
HarelM, line 123: Should this be attached to the document and not to the relevant html element? Won't this prevent the map from dragging?

src/components/LayerListItem.tsx
```diff
@@ -13,11 +13,19 @@ type DraggableLabelProps = {
   layerType: string
   dragAttributes?: React.HTMLAttributes<HTMLElement>
   dragListeners?: React.HTMLAttributes<HTMLElement>
+  onSelect?: () => void
```
HarelM: Can we make `onSelect` non optional to avoid the `?.()`?

HarelM: Also according to the definition below, this receives a layer index, doesn't it?

cypress/e2e/sidebar-resize.cy.ts
sank64: Good point — will refactor the Cypress test to use the existing MaputnikDriver helpers instead of raw cy.get() calls. Will push a follow-up commit.

src/components/AppLayout.tsx
sank64: Converting from a class component to a functional component means we use the useTranslation() hook instead of the withTranslation() HOC. The hook provides the same i18n instance directly — WithTranslation is only needed for the HOC pattern with class components. The functionality is identical, just using the modern React hooks API.

src/components/AppLayout.tsx
sank64: Sure! I will fix it

src/components/AppLayout.tsx
sank64: Agreed, will inline the style object directly in the JSX.

src/components/AppLayout.tsx
sank64: You're right — will wrap both resize handle titles with t() and add the corresponding keys to the translation files.

src/components/AppLayout.tsx
sank64: The document-level listeners only act when isDragging.current or isInnerDragging.current is true, which is set exclusively by mousedown on the resize handles. Normal map interactions (pan, zoom) are unaffected because those flags stay false. Attaching to document is intentional — it ensures we keep tracking the mouse even if the cursor moves outside the handle (which happens often during fast drags). Without this, releasing or moving outside the 5px handle strip would stall the resize

src/components/LayerListItem.tsx
sank64: Yes to both — will make onSelect required (non-optional) and update the type to onSelect: (layerIndex: number) => void to match the actual usage at the call site (onSelect={() => props.onLayerSelect(props.layerIndex)}). The ?.() can then be replaced with a direct call.

sank64:
Hi Harel, I've tried to implement all needed changes. I still get an error on [modals.cy.ts] (1 failing out of 37). I don't see how it is connected to the changes I made in the PR. 
- Kévin

src/components/AppLayout.tsx
HarelM: Can you add some AI generated translations?

HarelM:
Is there a react plugin that can facilitate for this? When I look at it now it looks like a lot of code and I'm wording if this can be solved using some plugin...?

src/components/AppLayout.tsx
sank64: I've added translations to the PR

sank64:
> Is there a react plugin that can facilitate for this? When I look at it now it looks like a lot of code and I'm wording if this can be solved using some plugin...?

I let you decide. We are using the code I PR to begin with. I wanted to help back the great maputnik team as a thanks for a great project. Unfortunately I won't have time to rewrite the PR or adjust it any more than I did.

HarelM:
I think this should be possible with: 
react-resizable-panels

It also has a good drag handle appearance for accessibility. 

I'll see if I can find the time for this. 
Thanks for taking the time to contribute this!

HarelM:
@KanishJebaMathewM if you want to push this forward too that'll be great.
I'd like to use an existing library for this instead of implementing it here.

KanishJebaMathewM:
@HarelM, I'll open a fresh PR for this using react-resizable-panels! Let's finish reviewing the other PRs first though, will pick this up after.

HarelM:
Sure thing, thanks!

HarelM:
- Supersede by #1870

PR #1791, opened 2026-04-08
fix: layer visibility tooltip and layer list state after delete (#1675, #1781)

### LINKED ISSUE #1675: Alt text is reversed for layer hide/show button
**Maputnik version**: 3.0.0
**Browser**: Brave
**OS**: macOS

**Description of the bug**:

Alt text is reversed when you hover over the show/hide toggle for a layer.

1. Add a layer
2. Hover over the show/hide button (👁️)
3. When the layer is hidden, you'll see the alt text "hide" when the action verb should be "show", and vice versa.

**Style file or style URL**:

n/a

**Screenshots**:

Image

Image

### LINKED ISSUE #1781: Layer deletion collapses all expanded groups
**Maputnik version**: main
**Browser**: Firefox
**OS**: Windows 11

**Description of the bug**:
Layer deletion collapses all expanded groups.

1. Open "OSM Liberty" style
2. Click "Expand"
3. Delete the background or any other layer
   Image
4. All the layer groups have collapsed
   Image

Other than being an anomaly, this behavior also makes it difficult to create example styles for Maplibre issue, as it complicates the process of cleaning-up real-life styles.

dakanggo:
## Summary

- **#1675** — The eye button’s `title` (hover tooltip) now matches the action that will run: **Hide** when the layer is visible, **Show** when it is hidden (`LayerListItem`).
- **#1781** — After deleting a layer, expanded prefix-groups stay expanded instead of snapping back to collapsed: collapse state is remapped when the layer list changes (`remapLayerListCollapsedGroups` in `layer.ts`, used from `LayerList`). Deleting a layer also keeps `selectedLayerIndex` / `selectedLayerOriginalId` in range so the list UI does not break (`App`).

## Test plan

- [ ] **#1675** — Add a layer, hover the visibility control: visible layer → tooltip says hide; hidden layer → tooltip says show; click still toggles visibility.
- [ ] **#1781** — Open e.g. OSM Liberty, click **Expand**, delete any layer: groups that were expanded remain expanded; selection remains sensible (no broken / stacked list items).

- Fixes maplibre/maputnik#1675  
- Fixes maplibre/maputnik#1781

### DIFF
--- a/cypress/e2e/layers-list.cy.ts
+++ b/cypress/e2e/layers-list.cy.ts
@@ -516,6 +516,43 @@ describe("layers list", () => {
     });
   });
 
+  /**
+   * Regression: deleting a layer must not reset expanded prefix-groups to collapsed
+   * (layer list items would stack / hide incorrectly).
+   */
+  describe("deletion with expanded prefix groups", () => {
+    beforeEach(() => {
+      when.modal.open();
+    });
+
+    it("keeps grouped layers visible after deleting another layer while groups were expanded", () => {
+      const gg = when.modal.fillLayers({
+        id: "gg",
+        type: "line",
+        layer: "example",
+      });
+      when.modal.open();
+      const gg2 = when.modal.fillLayers({
+        id: "gg-2",
+        type: "line",
+        layer: "example",
+      });
+      when.modal.open();
+      const solo = when.modal.fillLayers({
+        id: "solo",
+        type: "line",
+        layer: "example",
+      });
+
+      when.click("skip-target-layer-list");
+      then(get.elementByTestId(`layer-list-item:${gg2}`)).shouldBeVisible();
+
+      when.click(`layer-list-item:${solo}:delete`);
+      then(get.elementByTestId(`layer-list-item:${gg}`)).shouldBeVisible();
+      then(get.elementByTestId(`layer-list-item:${gg2}`)).shouldBeVisible();
+    });
+  });
+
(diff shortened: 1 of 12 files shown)


src/libs/layer.ts
```diff
@@ -89,3 +89,92 @@ export function findClosestCommonPrefix(layers: LayerSpecification[], idx: numbe
...
+}
+
+/** True if newIds is oldIds with exactly one id removed (order preserved). */
+export function isOneIdRemoved(oldIds: string[], newIds: string[]): boolean {
+  if (oldIds.length !== newIds.length + 1) {
+    return false;
+  }
+  let i = 0;
+  let j = 0;
+  while (i < oldIds.length && j < newIds.length) {
```
HarelM: Can this be converted to a for loop?

HarelM:
Looks good overall, thanks! 
Can you add an e2e test for the deletion scenario?

cypress/e2e/layers-list.cy.ts
```diff
@@ -516,6 +516,43 @@ describe("layers list", () => {
     });
   });
 
+  /**
```
HarelM, line 519: Use the test name instead of this comment please.

HarelM:
I think it will be better to split this PR to different PRs according to what they solve as I saw you added multiple sprite support which is not related to the previos issues.
So let's use a single PRs to fix a single issue.
Thanks for looking into all these issues!

PR #1850, opened 2026-05-02
feat: display current open filename in header bar

### LINKED ISSUE #1541: Maputnik does not show file name
__Maputnik version__: v3.0.0
__Browser__: Edge and Crome
__OS__: Windows

Maputnik separates the “style name” set under Style Settings and the "file name" when saving or loading a JSON file. These two are not automatically synchronized and only the style name is visible in the tab.

__Problem__: When I work with multiple Maputnik tabs, I easily lose track of which JSON file corresponds to which open editor window.

__Example workflow__:

Create new style → set style name = "map1"

Save file → filename defaults to "map1.json" → OK

Make changes → Save As → "map1_changed.json"

Open both files in separate Maputnik tabs

__Issue__: Both tabs now show the same style name ("map1") inside Maputnik, and the UI provides no indication of which file was originally loaded.

__Wish__: Ability to also see the file names in the UI.

KanishJebaMathewM:
## Fix: Display current open filename in header bar

Closes #1541

---

### Problem

Maputnik has two separate concepts:
- **Style name** — stored in JSON as the `name` field and shown in Style Settings
- **File name** — actual JSON filename on disk (e.g. `map1_changed.json`)

Previously, when a style file was opened, the filename was not retained in app state. As a result, users with multiple tabs open (same style name, different files) could not tell which tab belonged to which file.

---

### Fix

Introduced a single `currentFileName` value in app state that is:
- Set when a local file is opened via the Open dialog
- Updated on Save and Save As
- Shown in the top header area as a read-only file indicator
- Hidden when no local file is associated (fresh unsaved style or URL-loaded style)

Style name behavior in Style Settings is completely unchanged.

---

### Screenshot

image

> `map1.json` is now visible in the header bar, making it easy to identify which file is open in each tab.

---

### Files changed

| File | Change |
|---|---|
| `App.tsx` | Added `currentFileName` to `AppState`, initialized it, updated `openStyle()`, added `onSetCurrentFileName()`, wired props to toolbar and export modal |
| `ModalOpen.tsx` | Captures filename on local open paths and passes it to app state; URL-open keeps filename empty |
| `ModalExport.tsx` | Updates `currentFileName` on Save and Save As |
| `AppToolbar.tsx` | Renders read-only current filename indicator in header area |
| `_toolbar.scss` | Styling for filename indicator layout and visibility |

---

### Testing

- [x] Open a local JSON file — filename appears in header
- [x] Open two different local files in separate tabs — each tab shows its own filename
- [x] Save / Save As — header updates to the current filename
- [x] Fresh style (no local file loaded) — header shows no filename
- [x] URL-loaded style — header shows no filename
- [x] `npm run lint` passes
- [x] `npm run build` passes

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -121,6 +121,7 @@ type AppState = {
     codeEditor: boolean
   }
   fileHandle: FileSystemFileHandle | null
+  currentFileName?: string
 };
 
 export default class App extends React.Component<any, AppState> {
@@ -170,6 +171,7 @@ export default class App extends React.Component<any, AppState> {
         debugToolbox: false,
       },
       fileHandle: null,
+      currentFileName: undefined,
     };
 
     this.layerWatcher = new LayerWatcher({
@@ -611,8 +613,8 @@ export default class App extends React.Component<any, AppState> {
     }
   };
 
-  openStyle = (styleObj: StyleSpecificationWithId, fileHandle: FileSystemFileHandle | null) => {
-    this.setState({fileHandle: fileHandle});
+  openStyle = (styleObj: StyleSpecificationWithId, fileHandle: FileSystemFileHandle | null, currentFileName?: string) => {
+    this.setState({fileHandle: fileHandle, currentFileName});
     styleObj = this.setDefaultValues(styleObj);
     this.onStyleChanged(styleObj);
   };
@@ -849,6 +851,10 @@ export default class App extends React.Component<any, AppState> {
     this.setState({ fileHandle });
   };
 
+  onSetCurrentFileName = (currentFileName?: string) => {
+    this.setState({ currentFileName });
+  };
+
   onChangeOpenlayersDebug = (key: keyof AppState["openlayersDebugOptions"], value: boolean) => {
     this.setState({
       openlayersDebugOptions: {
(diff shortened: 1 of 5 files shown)


HarelM:
The provided screenshot doesn't look good. I suggest to add it to the browser tab, and maybe place it somewhere in the export modal or something like that. 
Note that you can also open urls and not files and this should be taken into consideration as well. 
Please add relevant end to end tests once we are on a good UX.

KanishJebaMathewM:
@HarelM I dont think adding the file which is currently opened in the browser tab will be providing a better user interface and do you think we should also display the currently opened url to the user as the urls may be longer and allocating a suitable space in the ux would be complicated 
image
This looks good for me, your views

HarelM:
The filename can be long and move the entire upper navbar, so placing the filename there is a no-go.
If you have a different place you think might be useful please suggest.

KanishJebaMathewM:
image
this looks much better

HarelM:
Please try a long file name...

KanishJebaMathewM:
https://github.com/user-attachments/assets/0f582214-0dba-4bd9-a139-a2af352ae6a7

The nav bar's length gets dynamically updated based on the file name

HarelM:
Can we leave that to be the last one? I'm not 100% aligned on the solution...

KanishJebaMathewM:
Ok sure

PR #1851, opened 2026-05-02
Fix/bounds field source editor

### LINKED ISSUE #1037: Styling: sources -> bounds is not reflected in the editor
**Description of the bug**:

Under the style specification, raster/vector sources can have an optional [bounds](https://maplibre.org/maplibre-style-spec/sources/#bounds) parameter. However, this value is not shown in the editor (hence it cannot be verified or modified in maputnik). This may cause confusion to users who are unaware of this issue.

1. Create or import a style with a data source that has bounds specified
2. Observe that the style is applied in the map, but not fully shown in the editor (i.e. the bounds value is not shown).

**Style file or style URL (partial)**:
```
  "sources": {
    "source": {
      "type": "vector",
      "minzoom": 0,
      "maxzoom": 17,
      "bounds": [140,20,150,30]
    }
  }
```

KanishJebaMathewM:
## Fix: Add bounds field to source editor UI

Closes #1037

---

### Problem

Raster and vector sources support the optional `bounds` parameter in the MapLibre style specification: a 4-value array `[west, south, east, north]` that defines the geographic bounding box of a source. Although Maputnik correctly reads and applies this value from JSON, it was never shown in the source editor UI. That meant users had to switch to the raw Code Editor to view or edit bounds.

---

### Fix

Added a `bounds` input field to both source editor types in `ModalSourcesTypeEditor.tsx`, following the same pattern used for `minzoom` and `maxzoom`:
- Four separate numeric inputs for west, south, east, north
- Optional field behavior: if `bounds` is not present in the JSON, the inputs stay empty
- Uses the existing default world bounds value `[-180, -85.051129, 180, 85.051129]`
- Applies to both `TileURLSourceEditor` and `TileJSONSourceEditor`

---

### Screenshot

> `bounds` values `[140, 20, 150, 30]` are now visible and editable in the source editor UI, right below Max Zoom

image

---

### Files changed

| File | Change |
|---|---|
| `ModalSourcesTypeEditor.tsx` | Added `bounds` to `TileURLSourceEditorProps` and rendered a `FieldArray` after `maxzoom` |
| `ModalSourcesTypeEditor.tsx` | Added `bounds` to `TileJSONSourceEditorProps` and rendered a `FieldArray` after the URL field |

---

### Testing

- [x] Load a style with `bounds` set — all four values appear correctly in the source editor
- [x] Load a style without `bounds` — fields remain empty and the UI does not crash
- [x] Edit bounds in the UI — changes are reflected in the JSON
- [x] `npm run lint` passes
- [x] `npm run build` passes


src/components/modals/ModalSourcesTypeEditor.tsx
```diff
@@ -37,6 +38,18 @@ class TileJSONSourceEditor extends React.Component<TileJSONSourceEditorProps> {
           url: url
         })}
       />
+      <FieldArray
+        label={t("Bounds")}
+        fieldSpec={latest.source_vector.bounds}
+        length={4}
+        type="number"
+        value={this.props.source.bounds || []}
```
HarelM: I'm not sure defaulting to the entire world is a good default, I would prefer an empty default and allow the user to add this only when relevant or present in the source definitions

HarelM:
Please revert filename changes in this PR as they are not relevant to this change. 
Please ad end to end test for this functionally.

src/components/modals/ModalSourcesTypeEditor.tsx
KanishJebaMathewM: You are right, emptied the default bounds so that the user can enter them

cypress/e2e/bounds-field.cy.ts
```diff
@@ -0,0 +1,173 @@
+import { MaputnikDriver } from "./maputnik-driver";
+
+describe("bounds field feature", () => {
+  const { beforeAndAfter, when, get, given, then } = new MaputnikDriver();
+  beforeAndAfter();
+
+  describe("when loading a style with bounds set", () => {
+    beforeEach(() => {
+      given.interceptAndMockResponse({
```
HarelM: This part should be in the driver.

cypress/e2e/bounds-field.cy.ts
```diff
@@ -0,0 +1,173 @@
...
+      when.click("modal:open.url.button");
+      when.wait(200);
+
+      // Open sources modal
+      when.click("nav:sources");
+      when.wait(200);
+    });
+
+    it("should display all four bounds values correctly in the source editor", () => {
+      // Navigate to see existing sources or click on a source to edit
```
HarelM: Please reduce comments as much as possible and use the test names for explaining the scenario.

cypress/e2e/bounds-field.cy.ts
```diff
@@ -0,0 +1,173 @@
+import { MaputnikDriver } from "./maputnik-driver";
+
+describe("bounds field feature", () => {
+  const { beforeAndAfter, when, get, given, then } = new MaputnikDriver();
+  beforeAndAfter();
+
+  describe("when loading a style with bounds set", () => {
+    beforeEach(() => {
```
HarelM: This is used for a single test as far as I can tell, so it might be better to place this code inside the test, or move it to a diver to allow better readability of the test. You can check out other tests to get some idea.

cypress/e2e/bounds-field.cy.ts
```diff
@@ -0,0 +1,173 @@
...
+    });
+  });
+
+  describe("when loading a style without bounds", () => {
+    beforeEach(() => {
+      when.setStyle("layer");
+      when.click("nav:sources");
+    });
+
+    it("should have empty bounds fields and not crash the UI", () => {
```
HarelM: Where do you test the UI crash?

cypress/e2e/bounds-field.cy.ts
```diff
@@ -0,0 +1,173 @@
...
+      when.select("modal:sources.add.scheme_type", "xyz");
+
+      // The bounds should remain empty (not default to world bounds)
+      then(
+        get.styleFromLocalStorage().then((style) => {
+          const source = style.sources[sourceId];
+          // Bounds should either not exist or be an empty array
+          return (source.bounds === undefined || source.bounds.length === 0);
+        })
+      ).should("equal", true);
```
HarelM: `should("equal", true);` looks like a mistake.

src/components/modals/ModalSourcesTypeEditor.tsx
```diff
@@ -37,6 +38,19 @@ class TileJSONSourceEditor extends React.Component<TileJSONSourceEditorProps> {
           url: url
         })}
       />
+      <FieldArray
+        label={t("Bounds")}
```
HarelM: Doesn't this needs translation?

src/components/modals/ModalSourcesTypeEditor.tsx
```diff
@@ -37,6 +38,19 @@ class TileJSONSourceEditor extends React.Component<TileJSONSourceEditorProps> {
           url: url
         })}
       />
+      <FieldArray
```
HarelM: I'm not sure bounds should be editable given that the source is using tilejson, much like other parameters of the source which are defined in the tilejson file...

KanishJebaMathewM:
give me some time

cypress/e2e/bounds-field.cy.ts
KanishJebaMathewM: sure

src/components/modals/ModalSourcesTypeEditor.tsx
KanishJebaMathewM: made the required translations in the latest commit

cypress/e2e/bounds-field.cy.ts
KanishJebaMathewM: The interceptAndMockResponse setup was now moved to the driver file (maputnik-driver.ts) and is now properly abstracted

cypress/e2e/bounds-field.cy.ts
KanishJebaMathewM: The setup code has been refactored and moved to the driver pattern following the project's test structure. The beforeEach block is no longer needed in this form

cypress/e2e/bounds-field.cy.ts
KanishJebaMathewM: Changed to .should("be.true") which is the correct Cypress assertion syntax.

src/components/modals/ModalSourcesTypeEditor.tsx
KanishJebaMathewM: Agreed! The bounds field has been removed from TileJSONSourceEditor.tsx and is only available in TileURLSourceEditor.tsx for tile sources where bounds are manually configured

cypress/e2e/bounds-field.cy.ts
KanishJebaMathewM: The crash validation is implicit - the test would fail if the UI crashed when handling empty bounds. I can add an explicit assertion to verify the modal remains visible for clarity.

cypress/e2e/maputnik-driver.ts
```diff
@@ -217,14 +224,36 @@ export class MaputnikDriver {
 
     setValueToPropertyArray: (selector: string, value: string) => {
       this.when.doWithin(selector, () => {
-        this.helper.get.element(".maputnik-array-block-content input").last().type("{selectall}"+value, {force: true });
+        const inputs = this.helper.get.element("input");
+        // Find the first empty input or use the first one
+        inputs.then(($inputs) => {
```
HarelM, line 236: It might be better to move this to a different method and have that specifically for a screen or something...

cypress/e2e/maputnik-driver.ts
```diff
@@ -217,14 +224,36 @@ export class MaputnikDriver {
...
+          let targetIndex = 0;
+          for (let i = 0; i < $inputs.length; i++) {
+            const input = $inputs[i] as HTMLInputElement;
+            if (!input.value) {
+              targetIndex = i;
+              break;
+            }
+          }
+          cy.wrap($inputs[targetIndex]).type("{selectall}" + value, { force: true });
```
HarelM: `cy.` should be part of the helper and not driver according to the architecture.

cypress/e2e/bounds-field.cy.ts
```diff
@@ -0,0 +1,126 @@
...
+      when.addValueToPropertyArray("modal:sources.add.bounds", "10.0");
+      when.addValueToPropertyArray("modal:sources.add.bounds", "10.0");
+      when.wait(100);
+
+      when.click("modal:sources.add.add_source");
+      when.wait(200);
+
+      then(
+        get.styleFromLocalStorage().then((style) => style.sources[sourceId].bounds)
+      ).should("deep.equal", [-10.0, -10.0, 10.0, 10.0]);
```
HarelM: Things like cypress specific `deep.equal` should not be present in this file.

cypress/e2e/bounds-field.cy.ts
HarelM, line 1: Shouldn't this be part of sources modal tests? Not sure it shouldn't have a different file...
IDK, your call...

HarelM:
Let's focus on this PR next.
I've added some comments, can you share the link with the updated version?

cypress/e2e/maputnik-driver.ts
KanishJebaMathewM: Will extract the input-finding logic into a separate helper method and have `setValueToPropertyArray` delegate to it

cypress/e2e/maputnik-driver.ts
KanishJebaMathewM: Will move the `cy.wrap(...).type(...)` call into the helper layer so the driver has no direct `cy.` calls

cypress/e2e/bounds-field.cy.ts
KanishJebaMathewM: Will remove the cypress-specific `deep.equal` assertion and move the bounds verification into the helper using the assertable pattern — will be updated in the next commit.

cypress/e2e/bounds-field.cy.ts
KanishJebaMathewM: Good point — will move the bounds tests into the sources modal test file to stay consistent with the existing structure

KanishJebaMathewM:
Please review the updated changes here: https://maputnik-bounds-field-source-editor.netlify.app/

cypress/e2e/maputnik-driver.ts
```diff
@@ -222,14 +229,36 @@ export class MaputnikDriver {
 
     setValueToPropertyArray: (selector: string, value: string) => {
       this.when.doWithin(selector, () => {
-        this.helper.get.element(".maputnik-array-block-content input").last().type("{selectall}"+value, {force: true });
+        const inputs = this.helper.get.element("input");
+        // Find the first empty input or use the first one
```
HarelM: Are you should this shouldn't be a different driver call? `setValueToPropertyArray` doesn't sound like it should find the first empty one...

HarelM:
There are still some open comments. Can you address those first please?

cypress/e2e/maputnik-driver.ts
KanishJebaMathewM: Good catch! Fixed `setValueToPropertyArray` to target index 0 directly, making it distinct from `addValueToPropertyArray`.

cypress/e2e/bounds-field.cy.ts
```diff
@@ -0,0 +1,126 @@
...
+      when.click("modal:sources.add.add_source");
+      when.wait(200);
+
+      const hasEmptyBounds = get.styleFromLocalStorage().then((style) => {
+        const source = style.sources[sourceId];
+        if (!source) return true;
+        return source.bounds === undefined || source.bounds.length === 0;
+      });
+
+      then(hasEmptyBounds).should("be.true");
```
HarelM: Sorry for nit-picking, this is still using cypress style check.
Use `shouldHaveValue` or `shouldEqual` instead.

cypress/e2e/bounds-field.cy.ts
```diff
@@ -0,0 +1,126 @@
...
+      when.click("nav:sources");
+      when.wait(200);
+
+      const sourceId = "tileurl-bounds-test";
+      when.setValue("modal:sources.add.source_id", sourceId);
+      when.select("modal:sources.add.source_type", "tile_vector");
+      when.select("modal:sources.add.scheme_type", "xyz");
+      when.wait(200);
+
+      when.setValueToPropertyArray("modal:sources.add.bounds", "-10.0");
```
HarelM: Wasn't this removed because tilejson doesn't support updating?

cypress/e2e/maputnik-driver.ts
```diff
@@ -222,14 +229,27 @@ export class MaputnikDriver {
...
       });
     },
 
     addValueToPropertyArray: (selector: string, value: string) => {
       this.when.doWithin(selector, () => {
-        this.helper.get.element(".maputnik-array-add-value").click({ force: true });
-        this.helper.get.element(".maputnik-array-block-content input").last().type("{selectall}"+value, {force: true });
+        const inputs = this.helper.get.element("input");
+        inputs.then(($inputs) => {
+          let targetIndex = 0;
```
HarelM, line 249: I still don't understand what this part of the code does.
Can you explain?
Or better yet move this part to a different method in the driver with the relevant documentation on when to use it vs the "regular" `addValueToPropertyArray`?
It's still not clear, sorry. Probably a cypress hack, but still...

cypress/e2e/bounds-field.cy.ts
KanishJebaMathewM: Changed from `should("be.true")` to `shouldEqual(true)` to use the custom Cypress helper assertion that matches the project's testing conventions.

cypress/e2e/bounds-field.cy.ts
KanishJebaMathewM: This test creates a new tile_vector source with XYZ scheme, not a TileJSON source. It sets bounds during source creation, which is different from updating bounds on an existing TileJSON source (where bounds come from the server and can't be updated). For new tile vector sources with XYZ scheme, bounds can be optionally set, so this test validates that feature works correctly in the UI.

cypress/e2e/maputnik-driver.ts
KanishJebaMathewM: Added comprehensive documentation comments explaining the difference between the two methods:

- `setValueToPropertyArray()` — Sets the first value in an array (index `0`), used for initialization.
- `addValueToPropertyArray()` — Finds the first empty input and fills it, used for sequential additions.

This pattern is necessary because array fields need initialization before additional values can be added.

The test demonstrates this by using `setValueToPropertyArray()` for the first bound value, followed by three `addValueToPropertyArray()` calls for the remaining bounds.

cypress/e2e/bounds-field.cy.ts
HarelM: This is still pending resolution...

KanishJebaMathewM: sorry , resolved in the latest commit

cypress/e2e/modals.cy.ts
```diff
@@ -120,6 +120,111 @@ describe("modals", () => {
...
+          when.click("modal:sources.add.add_source");
+          when.wait(200);
+
+          then(
+            get.styleFromLocalStorage().then((style) => style.sources[sourceId].bounds)
+          ).shouldDeepNestedInclude([-125.0, 25.0, -66.0, 49.0]);
+        });
+      });
+
+      describe("tilejson source with bounds", () => {
```
HarelM: Should this be tile url instead of tile json?

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@
 
 ### 🐞 Bug fixes
 
+- Moved bounds field Cypress coverage into the sources modal spec.
```
HarelM: Changelog should be the added bound feature, not this part and not under bug fixes...

KanishJebaMathewM: Moved bounds entry to **Features** and clarified it as: “Added bounds field and moved Cypress coverage into the sources modal spec.”

cypress/e2e/modals.cy.ts
KanishJebaMathewM: Renamed the test to `tile url source with bounds` to better match its intent (no functional changes).

HarelM (approved):

HarelM:
Looks like bounds tests are failing...
image

PR #1853, opened 2026-05-03
feat: add default OSM raster basemap for editor reference

### LINKED ISSUE #819: Add a default basemap
It would be great to have a default basemap just to localise our data.

The basemap can be OpenStreetMap (https://tile.openstreetmap.org/${z}/${x}/${y}.png) for example.
The basemap need to stay away from the MVT description, it wil not be in the json exported, it is only here to help the user to localise his data on the map and check the zoom level.

It always add a basemap that I need to drop befor exporting the json thus making the process to import+change+export more time consumming.

KanishJebaMathewM:
## Summary
Implements a runtime-only default basemap (CartoDB Light – no labels) 
that loads automatically in the map canvas, providing users with a 
clean visual reference for localizing their data and checking zoom 
levels.

Closes #819

## Problem
When opening Maputnik for the first time, the map canvas is completely 
blank with no geographical reference, making it hard to localize data 
and check zoom levels.

## Solution
A runtime-only basemap that:
- Loads automatically on editor initialization
- Only visible in the canvas — not in the layers panel
- Excluded from saved/exported JSON style files
- Uses CartoDB Light (no labels) to avoid visual noise

## Implementation
Modified `MapMaplibreGl.tsx`:
- `addDefaultBasemap()` — adds CartoDB Light raster source/layer to 
  the live map instance below all user layers
- Called on `style.load` event to persist across style changes
- Source/Layer IDs are namespaced to avoid conflicts with user styles

## Why CartoDB Light (no labels)?
- Minimal visual noise, won't compete with user layers
- Shows only basic geography at low zoom
- Progressive detail at higher zoom levels

image

### DIFF
--- a/src/components/MapMaplibreGl.tsx
+++ b/src/components/MapMaplibreGl.tsx
@@ -228,6 +228,13 @@ class MapMaplibreGlInternal extends React.Component<MapMaplibreGlInternalProps, 
     map.addControl(inspect);
 
     map.on("style.load", () => {
+      // Re-add the runtime-only basemap each time a new style is loaded.
+      try {
+        this.addDefaultBasemap(map);
+      } catch (e) {
+        // ignore
+      }
+
       this.setState({
         map,
         inspect,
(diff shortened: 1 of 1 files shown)


HarelM:
I'm not planning to review this code if you haven't. This is a garbage AI generated code. Closing.

KanishJebaMathewM:
Can fix the code, if the output is what you expected :)

HarelM:
I suggest to first make sure we properly define the problem and solution in the related issue before we move to implementation.

PR #1942, opened 2026-06-11
fix(security): bundle MapLibre RTL text plugin locally instead of loading from unpkg CDN

qdzsh:
## Summary

fix(security): 2 improvements across 2 files

## Problem

**Severity**: `High` | **File**: `src/libs/maplibre-rtl.ts:L3`

The app configures MapLibre to load the RTL text plugin directly from unpkg.com at runtime. This introduces a supply-chain risk: users must trust a remote script, and any compromise, outage, or DNS/man-in-the-middle issue can affect the application. Because the plugin is fetched dynamically, it is also not covered by the app's build-time dependency checks or integrity verification.

## Solution

Bundle the RTL plugin locally or serve it from a trusted, versioned application asset. If remote loading is unavoidable, pin the exact version and add integrity verification or a trusted self-hosted mirror.

## Changes

- `src/libs/maplibre-rtl.ts` (modified)
- `desktop/filewatch/filewatch.go` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

### DIFF
--- a/desktop/filewatch/filewatch.go
+++ b/desktop/filewatch/filewatch.go
@@ -12,7 +12,6 @@ import (
 var upgrader = websocket.Upgrader{
 	ReadBufferSize:  1024,
 	WriteBufferSize: 1024,
-	CheckOrigin:     func(r *http.Request) bool { return true },
 }
 
 func writer(ws *websocket.Conn, filename string) {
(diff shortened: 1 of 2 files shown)


qdzsh:
After looking into this more closely, I'm going to close it. Bundling the RTL text plugin locally doesn't build cleanly here: `setRTLTextPlugin()` expects a URL to a standalone worker script rather than a bundled module, and `@mapbox/mapbox-gl-rtl-text`'s `exports` map blocks importing the dist file directly, so the Vite build fails. Doing this properly would mean vendoring the worker script into the app's static assets and adjusting the build, which is really an architectural decision better left to you. The existing pinned-version CDN reference is a reasonable approach, so I don't think this change is worth the added complexity. Thanks!

PR #1988, opened 2026-07-08
Replace cypress with playwright

HarelM:
## Launch Checklist

This PR replaces cypress with playwright.
image

I hope this will make the end-to-end debugging and fixing easier due to the async await support of playwright which is missing in cypress.

This is a pure refactoring change.


HarelM:
I've reviewed the code myself several times to make sure it keeps the original architecture of test -> driver -> helper.
It seems to only fail when I broke something, but the tests did not fail for all the refactoring commits I did, so I think it's more stable.
This was assisted heavily by Claude code opus 4.8.

CommanderStorm (approved):

e2e/playwright-helper.ts
```diff
@@ -0,0 +1,414 @@
...
+
+/** Retries `assertion` until it stops throwing */
+async function retry(
+  assertion: () => Promise<void> | void,
+  timeout = 10000,
+  interval = 100
+): Promise<void> {
+  const start = Date.now();
+  let lastError: unknown;
+  for (;;) {
```
louwers: nit: `while (true)` is cleaner I think

birkskyum (approved):

e2e/playwright-helper.ts
HarelM: Fixed, thanks!

PR #1870, opened 2026-05-09
Sidebar resize react panels

KanishJebaMathewM:
Implements adjustable sidebar resizing using `react-resizable-panels` as a cleaner alternative to the custom implementation proposed in #1682.

This implementation:

* adds resizable sidebar width
* adds resizable split between layer list and properties drawer
* replaces custom resize logic with library-based panels
* simplifies resize handling and accessibility

The original PR #1682 was mainly used as UX/reference inspiration, but this implementation is rebuilt around `react-resizable-panels` for maintainability.

PR #2048, opened 2026-07-30
feat: make editor sidebar resizable and add dismiss close button to b…

raaz0000002:
1. Made the Editor Resizable & Scrollable:

Added a draggable splitter handle between the editor sidebar and the map.
Wrapped the editor panels in a horizontal scroll container so that if the sidebar is shrunk (reduced), you can scroll to see the layers list and details, increasing the map's visible viewport area.
Tied the bottom message panel's position dynamically to the sidebar's width to keep it aligned with the map.
Added a Bottom Panel Dismiss Button:

2. Added an "X" close button to the bottom message panel.
Allows dismissing the history/undo status messages (like "Undo removeLayer...") that appear on ctrl+z, clearing the panel.

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -927,6 +927,7 @@ export class App extends React.Component<any, AppState> {
       mapStyle={this.state.mapStyle}
       errors={this.state.errors}
       infos={this.state.infos}
+      onClearInfos={() => this.setState({ infos: [] })}
     /> : undefined;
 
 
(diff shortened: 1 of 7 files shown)


HarelM:
I think #1870 is a better solution to this problem from a UX perspective. Feel free to pick it up as it seems a bit stale...

HarelM:
No response, closing.

PR #2095, opened 2026-08-22
fix(desktop): open default browser on startup

### LINKED ISSUE #934: Desktop version should open browser
Let's make the desktop version open a browser window to the correct local url when started (http://localhost:8000).

reddynitish:
Fixes #934

## Summary

The Go desktop binary started its local server and printed the URL,
but never opened a browser automatically.

This PR:

- opens the default browser after the listener successfully binds
- uses the actual runtime URL
- uses a stdlib-only cross-platform launcher
- keeps browser-open failures non-fatal
- adds `--no-browser` for headless/Docker use
- leaves the normal Vite/web development flow unchanged

The listener is created before launching the browser so the browser
cannot race the server startup path.

## Platform behavior

- Windows: `rundll32 url.dll,FileProtocolHandler`
- macOS: `open`
- Linux: `xdg-open`

No shell command strings are used; arguments are passed directly through
`exec.Command`.

## Testing

- `go test ./...`
- `go vet ./...`
- `go build ./...`
- `git diff --check`
- repeated manual Windows startup verification (3 clean runs, confirmed the
  browser opened and hit the server, no duplicate launches)
- `--no-browser` verification (confirmed no launch attempt occurs)
- non-fatal launcher failure covered by a unit test that injects a
  nonexistent opener binary

macOS/Linux launch paths were not runtime-tested on this Windows machine —
they follow the standard `open`/`xdg-open` convention used across other Go
CLIs but are unverified here. The full packaged release flow (`gox`,
`go.rice`, `go-winres`) was not exercised locally.


desktop/browser_test.go
HarelM, line 1: This is a thin wrapper on browser package, I think it's not worth the tests...

HarelM (approved):

desktop/browser_test.go
reddynitish: Agreed — pushed e6ffc5b: removed both browser.go and browser_test.go, inlined the goroutine directly into maputnik.go's app.Action. It really was just a thin wrapper with no logic worth isolating or testing separately.

PR #2133, opened 2026-09-04
ci: keep the CodeQL actions on one version

johncarmack1984:
CodeQL has been failing on main since #2105: `init` is on v4.37.9 while `autobuild` and `analyze` are on v4.37.8, and the action rejects a config written by a newer version ("Loaded a configuration file for version '4.37.9', but running version '4.37.8'"). Dependabot bumps the three steps in separate pull requests, so #2119 and #2120 each still fail on their own and main has gone red at every partial bump (08-19, 08-28, today).

This moves the two steps to v4.37.9, the same commit dependabot picked in #2119 and #2120, and groups `github/codeql-action*` in dependabot so the three bump together from now on. Supersedes #2119 and #2120.

## Launch Checklist

 - [ ] Add an entry to `CHANGELOG.md` under the `## main` section (not applicable, CI only).

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -31,6 +31,10 @@ updates:
     directory: "/"
     schedule:
       interval: "daily"
+    groups:
+      codeql-action:
+        patterns:
+          - "github/codeql-action*"
     cooldown:
       default-days: 3
       # no semver support for github-actions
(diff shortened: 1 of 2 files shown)


HarelM (approved):
Thanks! Other repos probably need that too. Poor choice of architecture for this action I must say...

PR #2146, opened 2026-09-07
Fix layer group expansion state after deletion

### LINKED ISSUE #1781: Layer deletion collapses all expanded groups

NikitaMGrimm:
Deleting a layer changes the starting indices of later groups, so their stored expansion state no longer matches. Preserve that state through surviving layer IDs when rebuilding the group keys.

This keeps expanded groups open after deleting an earlier layer or the first layer of a group. Separate groups with the same prefix retain their own states; if deletion joins two groups, the merged group stays expanded when either was expanded.

- Fixes #1781.

Adds five deletion regression tests, a small style fixture, and a changelog entry.

Validation:

- All 33 layer-list E2E tests and 50 unit tests pass.
- Lint and the production build, including TypeScript checking, pass.
- Manually checked OSM Liberty: all 15 groups remain expanded after deleting the background layer.
- The full E2E suite reports 160 passing tests and one failure in `modals > open > upload` (`Could not get styleItem from localStorage`). The same failure reproduces with the original production code restored.

### DIFF
--- a/src/components/LayerList.tsx
+++ b/src/components/LayerList.tsx
@@ -198,20 +198,22 @@ class LayerListContainerInternal extends React.Component<LayerListContainerInter
       // group's state through its surviving layers, including its new first layer.
       const collapsedByLayer = new Map<string, boolean>();
       let idx = 0;
-      this.groupedLayers(prevProps.layers).forEach(layers => {
+      for (const layers of this.groupedLayers(prevProps.layers)) {
         const collapsed = this.isCollapsed(layerPrefix(layers[0].id), idx);
-        layers.forEach(layer => collapsedByLayer.set(layer.id, collapsed));
+        for (const layer of layers) {
+          collapsedByLayer.set(layer.id, collapsed);
+        }
         idx += layers.length;
-      });
+      }
 
       const collapsedGroups: {[key: string]: boolean} = {};
       idx = 0;
-      this.groupedLayers().forEach(layers => {
+      for (const layers of this.groupedLayers()) {
         const lookupKey = [layerPrefix(layers[0].id), idx].join("-");
         // If deletion joins two groups, keep the result expanded when either was.
         collapsedGroups[lookupKey] = layers.every(layer => collapsedByLayer.get(layer.id) !== false);
         idx += layers.length;
-      });
+      }
 
       if (!lodash.isEqual(collapsedGroups, this.state.collapsedGroups)) {
         this.setState({ collapsedGroups });
--- a/src/components/LayerList.tsx
+++ b/src/components/LayerList.tsx
@@ -106,13 +106,13 @@ class LayerListContainerInternal extends React.Component<LayerListContainerInter
     });
   };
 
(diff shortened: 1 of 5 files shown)


src/components/LayerList.tsx
```diff
@@ -193,6 +193,31 @@ class LayerListContainerInternal extends React.Component<LayerListContainerInter
   }
 
   componentDidUpdate (prevProps: LayerListContainerProps) {
+    if (prevProps.layers !== this.props.layers) {
+      // Group indices move when layers are deleted or reordered. Carry each
+      // group's state through its surviving layers, including its new first layer.
+      const collapsedByLayer = new Map<string, boolean>();
+      let idx = 0;
+      this.groupedLayers(prevProps.layers).forEach(layers => {
```
HarelM: nit: Can you use a regular for loop?

src/components/LayerList.tsx
```diff
@@ -106,13 +106,13 @@ class LayerListContainerInternal extends React.Component<LayerListContainerInter
     });
   };
 
-  groupedLayers(): (LayerSpecification & {key: string})[][] {
+  groupedLayers(layers = this.props.layers): (LayerSpecification & {key: string})[][] {
```
HarelM, line 109: It might be worth moving this to the lib folder to handle layer changes, maybe, I'm not sure.

NikitaMGrimm: I kept grouping in the component since it builds UI-specific keys

src/components/LayerList.tsx
```diff
@@ -193,6 +193,33 @@ class LayerListContainerInternal extends React.Component<LayerListContainerInter
   }
 
   componentDidUpdate (prevProps: LayerListContainerProps) {
+    if (prevProps.layers !== this.props.layers) {
+      // Group indices move when layers are deleted or reordered. Carry each
```
HarelM: Move comment into tsdoc

src/components/LayerList.tsx
```diff
@@ -193,6 +193,33 @@ class LayerListContainerInternal extends React.Component<LayerListContainerInter
   }
 
   componentDidUpdate (prevProps: LayerListContainerProps) {
+    if (prevProps.layers !== this.props.layers) {
```
HarelM: Can this if be negated to reduce indentation?

src/components/LayerList.tsx
HarelM: Or extracted into a new method at least.

HarelM (approved):

PR #1261, opened 2025-07-04
Refactor Field components with const arrow functions

louwers:
## Summary
- convert all remaining Field components to const arrow functions
- document the refactor in the CHANGELOG

## Testing
- `npm run lint`
- `npm run build`

------
https://chatgpt.com/codex/tasks/task_e_68684db1b6b88331837307a54e3f9dc1

PR #2153, opened 2026-09-09
Fix zoom property crash after deleting a stop

liuly0322:
## Launch Checklist

When a fixed paint value is converted to a zoom function, Maputnik creates exactly two stops.

image

Deleting either one of those two stops leaves a single remaining value. During the transition back to the ordinary field editor, `ZoomProperty` can render with `stops` undefined. It previously called `.map` unconditionally, causing the editor to crash.

image

This PR guards the stop list with optional chaining. The change only protects the render path and does not alter zoom-function conversion or value persistence.

Now, the remaining stop's value will be restored after deleting the other stop:

image

The regression test covers `circle-opacity`:

1. Convert a fixed value to a zoom function.
2. Change the second stop's value.
3. Delete the first of the two stops.
4. Verify that the ordinary field editor reappears.
5. Verify that the remaining stop's value is restored.
6. Edit the restored field and verify that the style updates successfully.

 - [ ] Link to related issues.

### DIFF
--- a/e2e/layer-editor.spec.ts
+++ b/e2e/layer-editor.spec.ts
@@ -289,6 +289,20 @@ describe("layer editor", () => {
         });
       });
 
+      test("should leave a plain editable value after deleting a stop", async () => {
+        await when.setFunctionStopValue("circle-radius", "Output value", 1, "0");
+        await when.deleteFunctionStop("circle-radius");
+        await then(get.elementByTestId("spec-field:circle-radius")).shouldBeVisible();
+        await then(get.styleFromLocalStorage()).shouldDeepNestedInclude({
+          layers: [{ id, paint: { "circle-radius": 0 } }],
+        });
+
+        await when.setValue("spec-field-input:circle-radius", "7");
+        await then(get.styleFromLocalStorage()).shouldDeepNestedInclude({
+          layers: [{ id, paint: { "circle-radius": 7 } }],
+        });
+      });
+
       test("should set the base", async () => {
         await when.setFunctionBase("circle-radius", "2");
         await then(get.styleFromLocalStorage()).shouldDeepNestedInclude({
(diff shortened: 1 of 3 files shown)


e2e/layer-editor.spec.ts
```diff
@@ -349,6 +349,28 @@ describe("layer editor", () => {
...
 
+    describe("zoom function collapse", () => {
+      beforeEach(async () => {
+        id = await when.modal.fillLayers({ type: "circle", layer: "example" });
+        await when.setValue("spec-field-input:circle-opacity", "0.4");
+        await when.makeZoomFunction("circle-opacity");
+        await when.setFunctionStopValue("circle-opacity", "Output value", 1, "0");
+      });
+
+      test("should restore an editable value after deleting the first stop", async () => {
```
HarelM: Are you sure the test is actually deleting a stop value? Does this test fail on main and passes now?

liuly0322: Yes, I've already used the test to find the bug was first introduced in #1261, it fails ever since then, and it passes now :)

HarelM: But the test doesn't read that it deletes a stop, maybe a utility method is needed in the helper/driver to make the test read better as removing a stop field?

e2e/layer-editor.spec.ts
```diff
@@ -446,6 +447,38 @@ describe("layer editor", () => {
       });
     });
 
+    for (const functionVariant of ["zoom", "data"] as const) {
```
HarelM: Keep tests DAMP, split those into several tests please.

liuly0322: done

e2e/layer-editor.spec.ts
```diff
@@ -408,6 +423,20 @@ describe("layer editor", () => {
         });
       });
 
+      test("should leave a plain editable value after deleting a stop", async () => {
```
HarelM: How is this test different than:
"should leave a plain editable value after deleting a stop"?

liuly0322: Identical, just to make sure both types of functions work. Should we remove one?

HarelM: Both end up in the same code you changed, right?
I think it's safe to remove one of them, unless it helps with coverage, which I'm not sure is the case here...

liuly0322: Removed the latter one, as it will not help with coverage. The `DataProperty`/`ZoomProperty` itself is already fully tested, these 2 new tests all follow the same `dataType` update path in `FieldFunction`, so one test is already enough.

HarelM (approved):

